PR #5, opened 2026-02-11
chore: :wrench: Update for ARM64 and mobile

moodyjmz:
Better builds for mobile and ARM64

Google CC doesn't run easily on ARM64 so I added a check to use native if possible, or fallback to Java

There were a few other niggly bits that error'd or built files in places where errors appeared when trying to access them

I see my formatting is slightly different, leave or change?

### DIFF
--- a/build/Gruntfile.js
+++ b/build/Gruntfile.js
@@ -74,12 +74,17 @@ module.exports = function(grunt) {
 		return configs;
 	}
 	function writeScripts(config, name) {
-		const develop = '../develop/sdkjs/';
+		const develop = process.env.BUILD_ROOT
+			? path.join(process.env.BUILD_ROOT, 'sdkjs', 'develop', 'sdkjs') + '/'
+			: '../develop/sdkjs/';
 		const fileName = 'scripts.js';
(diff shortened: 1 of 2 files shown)


rikled:
I don't know what exactly these develop files are but I think we should not have them in the default task and rather call `grunt default develop-compiled` if we want it? Otherwise it looks good. I like your formatting more

rikled (approved):

moodyjmz:
> I don't know what exactly these develop files are but I think we should not have them in the default task and rather call `grunt default develop-compiled` if we want it? Otherwise it looks good. I like your formatting more

This was a bit tricky as we need those files. However I overthought it and basically we just need the `--compile` flag added if we want compiled script output - which we would for it to work typically. So I changed web-apps/develop `make sdkjs` to call with this flag. Therefore we retain parity with the original, and also call the `compile` flag allowing the `writeScripts` part to compile the scripts needed to actually run the code in browser.

I got a bit confused with the paths through this - the solution is actually simple, I was trying to solve it inside the sdkjs, whereas the external caller sets the parameters.

PR #13, opened 2026-03-30
fix(license): remove stale proprietary code markers

tobias-weiss-ai-xr:
## Summary

Remove stale proprietary begin/end comment markers from sdkjs/cell/model/Serialize.js. The code between the markers is standard AGPL-licensed code — the markers were leftover upstream artifacts.

### DIFF
--- a/cell/model/Serialize.js
+++ b/cell/model/Serialize.js
@@ -12519,7 +12519,6 @@
                     return oThis.ReadPic(t,l, oDrawing);
                 });
             }
-            /** proprietary begin **/
             else if ( c_oSer_DrawingType.GraphicFrame == type )
             {
                 //todo удалить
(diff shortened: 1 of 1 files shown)


juliusknorr:
Also did some further digging around those:

The two removed lines originate from commit f21e542d5e52e86ae7e89caf603ff6187cbd0a65 by Igor.Zotov on 2013-12-05, with the message "для opensource версии" (Russian for "for the opensource version").

That commit added /** proprietary begin **/ and /** proprietary end **/ markers around the GraphicFrame deserialization block in Excel/model/Serialize.js (later renamed to cell/model/Serialize.js in commit 78f3145 in March 2016).

The markers were likely used to flag code that could be stripped for an open-source release, wrapping the GraphicFrame/chart-as-group parsing logic as "proprietary."

juliusknorr (approved):

PR #24, opened 2026-06-08
Fix invisible first column in mid-page continuous multi-column sections

waleedq:
## Problem

In a **continuous multi-column section that starts mid-page** — the classic academic-paper layout (full-width title + authors, then a 2-column body via a *continuous* section break) — the **first (left) column renders blank** on the page where the section begins. The text is still there (it stays selectable, and editing a visible part brings it back), it just isn't painted. The bug also persists in exported PDFs.

This is the long-standing two-column "invisible text" issue: ONLYOFFICE/DocumentServer#3642, ONLYOFFICE/DesktopEditors#2197 (tracked internally as #73801), reproduced by several users and still open.

A telling property reported by users (and confirmed here): it **toggles with the height of the content above** the section — adding/removing a line in the top area flips the left column on and off — which points at the section's mid-page start position rather than the content itself.

## Root cause

In `CDocument.prototype.Recalculate_PageColumn` (`word/Editor/Document.js`), the `recalcresult_PrevPage` handler re-enters a section by calling `currPageSection.Init()`, which **rebuilds every column of the section from scratch**, discarding the already-computed column content. Recalculation then **resumed mid-section** (`_ColumnIndex` was left at the last column), so every column *before* the resume point — column 0 — was wiped and never re-filled. Its draw range ended up empty (`Pos=0 / EndPos=-1`), so the painter skipped it while the content remained in the document model — hence "selectable but invisible".

This path is only taken when a continuous multi-column section **starts part-way down a page** *and* **overflows onto the next page** (the overflow is what triggers the `PrevPage` re-entry), which is exactly why it depends on the height of the content above the section.

## Fix

When re-entering a multi-column section via `PrevPage`, restart the re-fill from **column 0** (capturing the first column's start element *before* `Init()` wipes it), so the whole section is recomputed rather than resumed mid-way. ~5 lines, in the `PrevPage` branch only.

## Testing

- Adds `tests/word/document-calculation/multi-column-sections.html` — builds a full-width section followed by a continuous 2-column section that overflows onto the next page, and asserts column 0 is not empty. **Fails before this change, passes after.** (The shared test measurer reports a fixed line height regardless of font size, which flattens the vertical geometry and hides the bug, so the test installs a font-size-proportional measurer — see the test comment.)
- Existing `document-calculation` / `paragraph` suites still pass.
- Verified end-to-end: the real reproduction document now renders both columns (in editor and in exported PDF).

### DIFF
--- a/tests/runAll.js
+++ b/tests/runAll.js
@@ -50,6 +50,7 @@ const allTests = [
 	'word/custom-xml/custom-xml.html',
 	'word/document-calculation/floating-position/drawing.html',
 	'word/document-calculation/paragraph.html',
+	'word/document-calculation/multi-column-sections.html',
 	'word/document-calculation/table/correctBadTable.html',
 	'word/document-calculation/table/flowTablePosition.html',
 	'word/document-calculation/table/pageBreak.html',
--- a/tests/word/document-calculation/multi-column-sections.html
+++ b/tests/word/document-calculation/multi-column-sections.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+		<title>Multi-column section tests</title>
+		<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.6.0/jquery.min.js"></script>
+		<link type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/qunit/2.16.0/qunit.css" rel="stylesheet" media="screen" />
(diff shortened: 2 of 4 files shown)


juliusknorr:
Thanks for your pull request. The change for the mid-page fix looks fine to me. The other two comment coloring commits seem unrelated though. Could you please push them separately in a new PR?

waleedq:
Oh my bad, i am pushing to this branch by mistake, i will reset the branch and update you, thank you

waleedq:
@juliusknorr the branch is now reseted to the mid-page fix commits only

PR #29, opened 2026-06-09
build: default closure level to WHITESPACE_ONLY

juliusknorr:
Changes the default Closure Compiler `level` from `ADVANCED` to `WHITESPACE_ONLY`. Avoids aggressive renaming/inlining by default, keeping minified SDK output readable for debugging. Can still be overridden with `--level=ADVANCED`.


DmySyz (approved):

DmySyz:
Nice

DmySyz:
We could consider going back later, but for now that makes development easier I think.

PR #50, opened 2026-06-15
docs: add CLAUDE.md with codebase guidance for AI agents

MonaAghili:
## Summary

- Adds `CLAUDE.md` to give Claude Code and other AI agents accurate context about the sdkjs repo without needing to explore from scratch.

This file was generated entirely by AI (Claude Code). It serves as the starting point for AI agent context in sdkjs; future updates should extend it rather than replace it.

## What's documented

- Repo purpose and its place in the Euro-Office multi-repo workspace
- Directory layout (word, cell, slide, pdf, visio, common, build, configs, tests)
- Build system: Grunt + Google Closure Compiler, how to run debug vs release builds
- How to add new source files (configs JSON file-lists, concat order)
- Test harness: QUnit + node-qunit-puppeteer, no aggregate runner
- Code style gate: `python tests/code-style/check.py`


moodyjmz:
Nice — this is genuinely useful and I'm a 👍 to merge. The gotchas, naming conventions, and build workflow are all accurate (I checked them against the tree). Just a few cheap fixes worth doing before/right after merge:

- **Suite count is 66, not ~75** (counted the `.html` suites in `check-build.yml`).
- **Document the CI branch scope** — it runs on `fork`, `develop`, `release/**`, `hotfix/**` (push + PR), **not `main`**. The doc doesn't say so, and that's the one genuinely non-obvious, easy-to-trip-on fact. (Not a doc error — just a gap.)
- **Cross-reference `/DocumentServer/AGENTS.md`** and add a one-liner about the `AGENTS.md` vs `CLAUDE.md` naming collision. Other tools read different files — Cursor and Gemini/GPT-based agents read `AGENTS.md` (Cursor also has `.cursor/rules`), Claude reads `CLAUDE.md` — so keep one canonical source and have the others point at it rather than letting them drift.

Tiny nits, take or leave: `sdk.min` reads like a flat key (it's `sdk` → `min`); the `# CLAUDE.md` H1 is redundant. And FYI the `develop/sdkjs/*/scripts.js` wildcard is correct as written — don't let anyone "fix" it to a root-level file, that file doesn't exist.

The bigger "how do we keep this from drifting / where do future findings live" question is its own thing — split out to #51 so it doesn't block this. Keep this file small and orientation-only; long-tail findings go in the store described there.

Full review notes below for anyone (or any agent) who wants the detail. 👇

<details>
<summary>Pasteable review for an AI agent (Claude / Cursor)</summary>

````text
TASK: Apply review fixes to CLAUDE.md (PR #50, Euro-Office/sdkjs). Verify each against the
tree before editing; this is an orientation doc, keep it small.

MUST-FIX (factual, cheap):
1. Test-suite count: change "~75 suites" to "66 suites". Source of truth: count of `.html`
   suites enumerated in .github/workflows/check-build.yml (verified = 66).
2. Add CI branch scope. The workflow check-build.yml triggers on push + pull_request for
   branches: fork, develop, release/**, hotfix/**. It does NOT run on `main`. State this
   explicitly so an agent doesn't assume `main` PRs are CI-guarded. This is an omission to
   fix, not a wrong claim already present.
3. Cross-reference /DocumentServer/AGENTS.md (the workspace-root dev-environment doc: Docker
   stack, container exec, healthcheck). Add a note on the multi-runtime instruction-file
   collision: Cursor and Gemini/GPT-based agents read AGENTS.md (Cursor also reads
   .cursor/rules), Claude reads CLAUDE.md. Pick ONE canonical source and have the others
   reference it rather than duplicate, to prevent silent divergence across tools.

OPTIONAL NITS:
4. `sdk.min` / `sdk.common` notation reads like flat keys. Actual structure: a single top-level
   `sdk` object with sub-arrays min/common/desktop/mobile/exclude_mobile. Reword to avoid the
   "grep for sdk.min finds nothing" trap.
5. Remove the redundant `# CLAUDE.md` H1; start with the description.

DO NOT CHANGE (verified correct — guard against bad "fixes"):
- The `develop/sdkjs/*/scripts.js` wildcard is CORRECT. Only per-editor files exist
  (word/cell/slide/visio); there is NO root-level develop/sdkjs/scripts.js. `develop/` is
  gitignored build output.
- The LV-1050 gotcha, the no-ES-modules/concat-order constraint, the bracket-access-for-Closure
  rule, `make` vs `grunt`, and the TableId/co-editing model are all accurate and high-value.

SCOPE BOUNDARY:
- Keep must-know invariants (CI scope, LV-1050, "never hand-edit generated bundles") IN this
  always-loaded file — do not push them into a lazily-retrieved store.
- The mechanism for long-tail findings (frontmatter docs + derived index + local vector DB
  cache) is tracked in issue #51, NOT this PR. Just have CLAUDE.md name that store as the
  canonical home for future findings once #51 lands.
````

</details>

**Review by moodyjmz — CHANGES_REQUESTED** [review]
See comments

**Comment by MonaAghili** [comment]
@moodyjmz
The changes have been applied.

Over the coming days, I’ll add more rules that I have in mind. As Jan suggested, those changes should be submitted in separate PRs.

### Applied changes

* Corrected the test suite count from approximately 75 to 66 (verified against `check-build.yml`).
* Documented the CI branch scope as `fork`, `develop`, `release**`, and `hotfix**` only (excluding `main`).
* Added a **"Multi-tool instruction files"** section covering `AGENTS.md` vs. `CLAUDE.md`, including collision handling and the canonical source rule, with a cross-reference to `/DocumentServer/AGENTS.md`.
* Clarified that the SDK configuration uses `sdk` → `min` / `common` sub-arrays rather than flat keys.
* Replaced the redundant `# CLAUDE.md` H1 heading with `# sdkjs — Euro-Office JavaScript SDK`.
* Added a pointer to issue `#51` indicating where future findings will be stored.

**Comment by juliusknorr** [comment]
Very nice, also generally think this is 
> Corrected the test suite count from approximately 75 to 66 (verified against check-build.yml).

Maybe we can prevent such hardcoded numbers in general?

> Cross-reference /DocumentServer/AGENTS.md and add a one-liner about the AGENTS.md vs CLAUDE.md naming collision. Other tools read different files — Cursor and Gemini/GPT-based agents read AGENTS.md (Cursor also has .cursor/rules), Claude reads CLAUDE.md — so keep one canonical source and have the others point at it rather than letting them drift.

General pattern in other open source projects seems to be to have the instructions in AGENTS.md and claude just referring to that. I would vote to try to make this accessible to a wider range of coding agents easily.

As Jan mentioned you would regenerate the file more often, would it make sense to store the prompt to generate them somewhere as well so it an be fine tuned over time?

**Review by juliusknorr — APPROVED** [review]
Just a few questions but fine from my side as well

**Comment by MonaAghili** [comment]
@juliusknorr 
most agents now read both `AGENTS.md` and `CLAUDE.md`, but AGENTS.md has wider baseline support across non-Claude tools.
We can agree on a standard, since our team is planning to add it to every repository.

**Comment by MonaAghili** [comment]
> As Jan mentioned you would regenerate the file more often, would it make sense to store the prompt to generate them somewhere as well so it an be fine tuned over time?

Regarding to storing the generation prompt

Yes worth doing. Common approaches:

- `.claude/prompts/generate-agents-md.md` a file containing the prompt, versioned in git alongside the output. When the doc drifts, you edit the prompt and regenerate.
- A comment block at the top of `CLAUDE.md`: `<!-- Generated by: see .claude/prompts/generate-agents-md.md -->`

What do you think? @juliusknorr @moodyjmz

**Comment by MonaAghili** [comment]
> Very nice, also generally think this is
> 
> > Corrected the test suite count from approximately 75 to 66 (verified against check-build.yml).
>  Maybe we can prevent such hardcoded numbers in general?

So are you suggesting we remove the suite count entirely or replace it with a pointer to the source of truth?

Option: replace with:
CI enumerates suites explicitly in `.github/workflows/check-build.yml`. If you have shell access: `grep -c "\.html" .github/workflows/check-build.yml` no hardcoded number to go stale.

**Comment by juliusknorr** [comment]
Sounds good to me. 

Claude Code does not read AGENTS.md by default i think, at least in my tests it did not pick this up, only after i stated that in my global CLAUDE.md file. I would settle on AGENTS.md and prefer to keep things generic, but also would not block having claude specific files in the repo.

**Comment by moodyjmz** [comment]
> Sounds good to me.
> 
> Claude Code does not read AGENTS.md by default i think, at least in my tests it did not pick this up, only after i stated that in my global CLAUDE.md file. I would settle on AGENTS.md and prefer to keep things generic, but also would not block having claude specific files in the repo.

I think we should have a reference to it in CLAUDE.md and whatever else. ie, we have one file with details and for major AI's we use their protocol to ensure they read the file.

**Comment by moodyjmz** [comment]
Building on "use each tool's protocol to point at the canonical file" — for Claude Code that protocol is an `@import`, so the stub can literally be one line:

```
@AGENTS.md
```

Claude Code expands that on load, so we keep a single source of truth in `AGENTS.md` (generic — works for Cursor/Gemini/GPT out of the box) and Claude picks it up automatically, with zero duplication and nothing to keep in sync. If we ever do want a few Claude-only lines, same trick: `@AGENTS.md` at the top, extras below.

One caveat for the follow-up PR: we'd then have two `AGENTS.md` — the workspace `/DocumentServer/AGENTS.md` (env/Docker) and a new per-repo one (sdkjs specifics). That's fine as long as we're explicit they're different scopes and the repo one points up for environment stuff — otherwise we just move the divergence problem up a level.

File: CLAUDE.md
Comment (Alex-Arsys) [inline CLAUDE.md:1]: Hey — the @AGENTS.md import pattern makes sense for keeping the operational layer (build, run, commit) as a single source of truth. One thing worth discussing though:

AGENTS.md covers how to work in the repo. CLAUDE.md has a different job: it's where the agent learns how to write code for the repo — the conventions and constraints that aren't discoverable from reading the code cold. For sdkjs that layer isn't trivial: the asc_* / get_/put_ / c_oAsc* naming conventions, the IIFE module pattern, the fact that every document mutation requires a change object for undo and co-editing to work, the file ordering in configs/*.json as actual execution order. None of that is in AGENTS.md, and an agent without it can produce plausible-looking code that quietly violates these invariants.

Genuinely curious about the thinking here — if CLAUDE.md right now just imports AGENTS.md and doesn't add anything on top, is it actually pulling its weight as a separate file? It feels like we're carrying the maintenance overhead of two files while only getting the informational value of one. Maybe the plan is to grow it with code-specific content over time, which would make total sense — just wanted to check whether that's the direction or if the current state is intentional!

@juliusknorr @moodyjmz

File: CLAUDE.md [hunk 1]
```diff
@@ -0,0 +1,3 @@
+# sdkjs
+
+See [AGENTS.md](AGENTS.md) for codebase guidance.
```
moodyjmz: This needs to be `@AGENTS.md` on its own line - I think that is all that is needed in this file

moodyjmz:
This needs x referencing `/DocumentServer/AGENTS.md`

CLAUDE.md
moodyjmz: Claude.md is for claude, AGENTS is for everything else, but shared with claude. To avoid duplication, claude.md imports agents.md. it is one line. It removes maintenance overhead.

moodyjmz:
To save pain, we should mention about docker:

For Docker dev environment setup (running the full server stack), see /DocumentServer/AGENTS.md

moodyjmz (changes requested):
couple of comments

moodyjmz (approved):
🤖 **Reviewed by Claude Code**

Both open items from the previous pass are resolved:
- `CLAUDE.md` → `@AGENTS.md` (Claude Code will now auto-load it)
- Cross-reference to `/DocumentServer/AGENTS.md` added in the right place

Content verified against the tree. LGTM.

PR #62, opened 2026-07-15
fix: switch data URI to blob URI

DmySyz:


moodyjmz:
# Code Review: PR #62 — fix: switch data URI to blob URI

**TL;DR:** Don't merge yet. The idea is right — swap the old `data:` URI download for a `Blob` URL, since giant data URIs choke some mobile browsers — but the patch as written can crash instead of silently doing nothing, and it hand-rolls a worse version of a download helper the codebase already has everywhere else. Two fixes needed: (1) bail out early if there's no image data instead of letting it throw, (2) call the existing `AscCommon.DownloadFileFromBytes` helper instead of reimplementing anchor/Blob/cleanup logic inline. No security concern — the image data never comes from an attacker, just from the editor's own canvas.

<details>
<summary><strong>Full technical review</strong></summary>

## Overview

`baseEditorsApi.prototype.asc_SaveDrawingAsPicture` used to build a `data:` URI from `oController.getImageDataForSaving(...)`'s `.src` and set it directly as an `<a>` anchor's `href`. This PR decodes the base64 payload to a `Uint8Array` via `atob()` + a `charCodeAt` loop, wraps it in a `Blob`, creates an object URL, sets that as the anchor's `href`, and revokes it after a 10-second `setTimeout`.

Reviewed with an independent cold second pass and a separate adversarial refutation pass (both spawned blind). Both converged on the same two core issues with no refutations.

## Findings

### 1. HIGH — New crash on empty `src`, not present in the old code

```js
let sMime = sSrc.split(',')[0].split(':')[1].split(';')[0];
```
When `sSrc === ""`: `.split(':')[1]` is `undefined`, and `undefined.split(';')` throws `TypeError`. Verified live in node. No try/catch anywhere on this path. Old code degraded silently on the same input (`a.href = ""`). Note the `if(oImageData)` guard doesn't help — `oImageData` is the wrapper object `{src, width, height, bounds}`, truthy even when `src === ""`.

`""` is a documented, reachable return value, not a hypothetical:
- `common/Drawings/CommonController.js:9832-9834` — when `NATIVE_EDITOR_ENJINE` is set (native/mobile builds), `getSelectionImageData()` unconditionally returns `""` for any multi-shape selection. Given this PR targets a mobile download issue, **this is likely the most-hit case on the exact platform being fixed**.
- `common/Drawings/CommonController.js:9827-9831` — tainted-canvas `SecurityError` (cross-origin image without CORS) is caught and falls back to `""`.
- `common/Drawings/Format/GraphicObjectBase.js:3782, 3789-3792, 3809-3813` — `""` for Notes/NotesMaster-parented shapes, non-finite/zero-extent shapes, or failed conversion with no cached data.

### 2. HIGH — Duplicates an existing, more robust download helper

`AscCommon.DownloadFileFromBytes` (`common/editorscommon.js:2433-2470`) already does bytes → Blob → object URL → anchor → click → cleanup, and is the house pattern used in `word/api.js:9491`, `cell/api.js:1555`, `slide/api.js:8504`, `visio/api.js:957`. It has **four** things this PR's inline version lacks — three across browsers, and one that hits the exact platform this PR targets:
- `tempLink.classList.add("external")` (`common/editorscommon.js:2450`) — **the mobile-critical one.** Prevents the framework7-react click hook (`web-apps/vendor/framework7-react/.../clicks.js`) from intercepting the anchor click. Without it, the download can silently do nothing on the mobile editor — i.e. the very platform this fix is for. The inline version omits it entirely.
- `document.body.appendChild(tempLink)` before `.click()` — more robust across browsers.
- A Safari popup-blocker `_blank` fallback — **not cosmetic**, since this PR is fixing a mobile issue.
- A `200ms` revoke delay (commented, tied to a specific webkit bug) vs. this PR's unexplained `10000ms`.

`getImageFileFromDataURL` (`common/editorscommon.js:2472-2477`) is an even closer existing match, using `AscCommon.Base64.decode` instead of hand-rolled `atob`/`charCodeAt`. Caveat: it has the same unguarded-empty-input problem, so finding 1's guard is still required either way — reuse alone doesn't fully fix it.

**Recommended fix:** guard on falsy `src`, decode via the existing `Base64.decode` pattern, call `AscCommon.DownloadFileFromBytes(bytes, filename, mime)` — but see finding 3 first.

### 3. MEDIUM — Findings 2 and 3 are in tension; resolve before shipping

`word/api.js:9469`: `//DownloadFileFromBytes has bug on Chrome on iOS https://github.com/kennethjiang/js-file-download/issues/72`. This directly qualifies finding 2's recommendation — you can't blindly tell the author to delegate to `DownloadFileFromBytes` when the helper has a documented Chrome-on-iOS bug and this is a mobile fix.

The resolving fact: the inline code uses the **same mechanism** as `DownloadFileFromBytes` (Blob → `createObjectURL` → anchor → click → revoke). So either the iOS bug hits both — in which case this PR doesn't actually fix the reported mobile problem — or the fix lives in one of the *differences* (appendChild / `external` class / target), in which case reuse is strictly better and finding 2 wins outright. Worth noting `word/api.js` still calls `DownloadFileFromBytes` for HTML/DOCX in production despite that noted bug, which suggests the bug is narrow and reuse is the right call.

**The one question that resolves everything:** @DmySyz — what did you test this on, and did the original bug reproduce when you tried plain `AscCommon.DownloadFileFromBytes`? If you never tried the house helper, finding 2 stands. If you tried it and it failed on-device, tell us which difference made your version work so we keep it.

### 4. LOW — Performance is a non-issue

The manual decode loop is `O(n)`, same complexity as `atob`/`Base64.decode` alone. Blob/ObjectURL is directionally correct for large images vs. giant data-URI strings. Only complaint is duplication (finding 2), not speed.

## Security

No injection/XSS risk — `oImageData.src` always originates from the editor's own `canvas.toDataURL()` / internal converters, never attacker-controlled text. The only security-adjacent issue is the availability bug in finding 1 (uncaught exception on reachable input).

## Summary

**Merge?** No — request changes.

**Must-fix before merge:**
1. Guard against falsy/empty `src` before any string operations.
2. Decode via `Base64.decode` and delegate to `AscCommon.DownloadFileFromBytes` instead of reimplementing anchor/Blob/revoke handling — this also restores the framework7 `external`-class handling the inline version drops.
3. Confirm on-device whether `DownloadFileFromBytes` reproduces the bug this PR set out to fix (Chrome-iOS caveat, `word/api.js:9469`) before choosing reuse vs. keeping a hand-rolled variant.

**Nit:** 10s revoke delay is an unexplained magic number vs. the 200ms used elsewhere for the same op — moot if fix #2 is applied.

**Noted, not flagged:** all new bindings (`sSrc`, `sExt`, `sMime`, `sBinary`, `aBytes`, `oBlobUrl`, `a`) use `let` despite never being reassigned. Not raised as a finding — `common/apiBase.js` already runs 109 `let` vs. 69 `const`, and `.eslintrc.yaml` has no `prefer-const` rule, so this is consistent with existing style, not a regression.

## Review methodology

Independent cold second pass (fresh agent, blind) + adversarial refutation pass (instructed to disprove the findings with code evidence). Both confirmed findings 1–3 with no refutations. Findings re-grounded against source on a second read: crash reproduced in node, helper diffs and the framework7 `external`-class handling confirmed at `common/editorscommon.js:2450`, and the Chrome-iOS caveat confirmed verbatim at `word/api.js:9469`.

</details>

PR #21, opened 2026-05-19
Update from master branch in upstream repository

rikled:
This pull request updates the `main` branch with changes from the `master` branch in the upstream repository.

PR #65, opened 2026-07-23
build(sdkjs): fix webpack migration bundle bloat, CI, and coverage gaps

MonaAghili:
## Summary

Fixes four bugs/gaps found in review of the Grunt→webpack migration (`feature/migration-to-webpack-sdkjs`), scoped strictly to build tooling / CI, plus one small related dead-code removal in application code (see "Out of scope" below) — the other three application-code findings are held for a separate PR.

### 1. Terser was duplicating every file's license header into the bundle
`build/webpack.sdk.factory.mjs` configured Terser to preserve comments matching `/AGPL|Copyright|Ascensio|License/i`, intending to keep just the single license banner injected by `BannerPlugin`. But every one of the ~400+ concatenated source files carries an identical AGPL/Copyright header, so the regex preserved all of them instead of just the one banner.

Verified before/after on the word bundle: **307 → 1** occurrences of the copyright text in `sdk-all.js`, banner still present and intact.

Fix: added a unique sentinel (`@@license-banner@@`) to `build/license.header` and match only that in Terser's `format.comments`.

Note: `BannerPlugin` injects the banner *before* Terser runs (webpack stage `ADDITIONS` (-100) vs Terser's `OPTIMIZE_SIZE` (400)), so a naive `comments: false` strips the banner too — confirmed by building and grepping the output. The sentinel is required, not optional.

### 2. Desktop/mobile silently lost their lighter minification tier
Old `build-desktop.bat`/`build-mobile.command` ran Closure's `WHITESPACE_ONLY`; the new pipeline funneled all platforms through the same Terser `compress` pass, meaning desktop/mobile now get more aggressive minification than before with no explicit decision behind it.

Fix: `compress` is now `false` for `SDK_PLATFORM=desktop|mobile`, restoring the old lighter tier. Verified by building with `SDK_PLATFORM=desktop`: output is larger than the web build (2.44 MiB vs 2.29 MiB min chunk), confirming the lighter pass is active.

### 3. `check-build.yml` PR trigger filter was dropped
`pull_request.branches` (`fork`, `develop`, `release/**`, `hotfix/**`) was removed during the CI-runner rewrite, so the workflow now runs on PRs targeting any branch. Restored.

### 4. Only `word` had a compiled-bundle regression test
The only CI check that runs QUnit against the actual webpack-built output (`tests/common/api/api.html`, under `COMPILED=1`) covered `word` only. `cell`/`slide`/`visio` had no equivalent — a bundle-specific regression (bad concat order, a stripped directive, etc.) in any of those three could pass CI undetected.

Verified the four editors' bootstrap ("min") chunks are genuinely distinct file sets (word: 42 files incl. `pdf/api.js`; cell: 38 files incl. `cell/api.js`, `cell/model/CollaborativeEditing.js`; etc.) — so this isn't redundant coverage.

Fix: added `tests/common/api/api-cell.html`, `api-slide.html`, `api-visio.html`, reusing the existing `api.js` test (it only exercises editor-agnostic `AscCommon.*` APIs, not `AscWord`/`AscCommonExcel`, so it's safe against the documented `COMPILED=1` limitation), wired into the same CI step.

## Also documented (no functional change)
- `output.publicPath` is intentionally unset today (no code-splitting/`import()` in this config); added a comment so the next person adding `splitChunks` knows to set it then.

## Test plan
- [x] `npm test --prefix build` — 32/32 build-tooling unit tests pass
- [x] `npm run build --prefix build` — all four editors (word/cell/slide/visio) build successfully
- [x] `SDK_PLATFORM=desktop npm run build --prefix build` — builds successfully, confirms lighter minify branch is exercised
- [x] `COMPILED=1 npm run develop --prefix build` — regenerates all four `scripts.js`, each correctly bootstrap-only
- [x] New smoke-test HTML files verified to resolve to real on-disk bundles
- [x] `check-build.yml` YAML validated
- [ ] CI run on this PR (Puppeteer/QUnit execution not run locally — no root `node_modules` in this environment)

## Out of scope
Three other findings from the same review are **not** included here — a spreadsheet checkbox feature, dead code in `TableId.js`, and a branding string change — since they're unrelated application-code changes riding along in the migration branch and belong in a separate PR reviewed on their own merits.

One related fix **is** included: `common/Drawings/Format/Controls.js` had a duplicate `CControl`/`CControlPr`/`FormControlPr` factory registration that `TableId.js`'s `private_InitFactoryClass()` was silently clobbering anyway (it resets `m_oFactoryClass` and re-registers the same three classes later, at editor-construction time). Removing the dead registration is low-risk and directly relates to the webpack chunk-ordering assumption the original comment was wrong about, so it's kept with this PR rather than split out. Verified against `forms.html`/`complexForm.html`.


moodyjmz:
## TL;DR

**Approve with reservations.** Build-tooling-only PR (no editor/application code touched); the four bugs it claims to fix are real fixes. One concrete defect ships in every output file and should be fixed before merge: the `@@license-banner@@` sentinel added to distinguish the license banner from per-file headers is never stripped, so it's now visible in every shipped bundle and asset. I also traced through the new cross-chunk global-sharing model empirically (built the real webpack config, no fabricated claims) — it's *not* broken today, but it depends on an undocumented webpack optimization that a future change (code-splitting, an external, `import()`) would silently break, with no test currently able to catch it. `build/DEVELOPER-GUIDE.md` is referenced four times and doesn't exist. Details below, ranked by severity.

<details>
<summary><strong>Full review</strong></summary>

### 1. Every shipped file's license banner literally contains the text `@@license-banner@@` — confirmed, trivial fix

`build/license.header:1` adds the sentinel `/* @@license-banner@@` so Terser's `format.comments` regex can distinguish the single injected license banner from the ~400 per-source-file AGPL headers that also match `/AGPL|Copyright|Ascensio|License/i` (this is bug #1 the PR fixes — good catch on the original problem). But nothing strips the sentinel back out before the text ships:

- `build/webpack.sdk.factory.mjs` builds `licenseText` via four `.replace()` calls (`@@AppCopyright`, `@@PublisherUrl`, `@@Version`, `@@Build`) and passes it verbatim into `webpack.BannerPlugin({ banner: licenseText, raw: true })`. Terser is then told to *preserve* comments matching `/@@license-banner@@/` — i.e. keep it verbatim, sentinel included.
- `build/scripts/deploy-assets.cjs` does the same four replaces, then does `licenseText + '\n' + code` for every individually-deployed JS file (`device_scale.js`, `path-boolean-min.js`, etc.).

A whole-tree grep for `license-banner` returns exactly two hits (the header source and the Terser preserve-regex) — no strip step anywhere. Every `sdk-all.js`, `sdk-all-min.js`, and every individually-deployed JS asset will visibly show `@@license-banner@@` in its header comment, forever.

**Fix:** strip the sentinel out of `licenseText` after it's built (before injecting), and keep it only in the separate string Terser matches against — don't conflate "the marker used to find the banner" with "the text that ships."

### 2. The cross-chunk global-sharing model rests on an undocumented, untested webpack optimization — not a bug today, but a landmine

sdkjs relies on ~287 places (`cell/view/EventsController.js`, `word/Drawing/DrawingDocument.js`, `pdf/src/viewer.js`, etc.) reading a bare `GlobalSkin` global with **zero** `window.GlobalSkin` assignment anywhere in the codebase — by design (`common/skin.js`'s own comment: *"не скрываем переменные, скин используется напрямую в sdk-all.js"* — "we don't hide the variables, the skin is used directly in sdk-all.js"). This only works because a top-level `var` in a plain `<script>` tag becomes a `window` property automatically.

I flagged this initially as broken, on the theory that webpack always wraps module code in a function regardless of `output.iife`. I verified this empirically by building the actual config with the real webpack (borrowed from `../web-apps/build/node_modules`, no new downloads) — that premise turned out to be wrong for this specific setup: with `iife:false` and a single module with no imports (exactly what `sdk-concat-loader` produces for each chunk), webpack **inlines the code at true top level with no wrapper**, so bare `var`s still land on `window`, matching the old Grunt output exactly.

So it works — but only because of that single-module/no-runtime special case. The moment this config gains `import()`, `splitChunks`, or an external, webpack switches to its wrapped bootstrap form and all 287 bare-global references would silently break at runtime (`TypeError: Cannot read properties of undefined`), with a green build and no warning. Nothing in the code currently connects this scoping dependency to the `publicPath`/no-code-splitting comment that's already sitting right next to it in `webpack.sdk.factory.mjs`.

**Suggestion:** add a comment on `output.iife: false` (or near it) spelling out that correctness here depends on staying single-module/no-runtime per chunk, so the next person touching this config doesn't unknowingly break it.

### 3. The CI check meant to catch exactly that kind of regression never actually ran on this PR

`.github/workflows/check-build.yml`'s `pull_request.branches` filter (`fork`/`develop`/`release/**`/`hotfix/**`) doesn't include `main` — and this PR's base branch is `main`. So neither the `unit-tests` job nor the new `build-test` job (which includes the `COMPILED=1` QUnit smoke test added here as bug-fix #4) ran in CI for this PR. The test-plan checkbox for it is unchecked in the PR body ("CI run on this PR — Puppeteer/QUnit execution not run locally, no root node_modules in this environment").

Separately: even if it had run, `tests/common/api/api.js` only exercises `AscCommon.*`/`AscUrlType.*` (explicit-namespace APIs) — it never touches `GlobalSkin` or any other bare-global-style symbol, so it wouldn't have caught the scenario in finding #2 anyway.

Net effect: the coverage gap this PR claims to close (bug #4) is real and a good fix, but it's narrower than the description implies, and unverified for this specific submission. Recommend actually running `COMPILED=1 npm run --prefix build develop` + the new `api-cell/slide/visio.html` suites locally before merge, rather than relying on the (currently non-triggering) CI job.

### 4. `build/DEVELOPER-GUIDE.md` is referenced four times but doesn't exist anywhere in the PR or on `main`

`AGENTS.md` (twice), `build/Readme.md` (once), and — this one's worse — `build/scripts/build-pipeline.cjs`, which prints the path in a user-facing error message shown to anyone who passes a stale Grunt CLI flag. Either add the file or drop the references.

### 5. Minor: Babel transpile cache key has two staleness gaps (low severity, dev-loop impact only)

`build/loaders/sdk-concat.cjs`'s `BABEL_OPTIONS_KEY` is a hand-maintained JSON literal mirroring the actual options passed to `transpileToES5`, despite the adjacent comment implying it's derived from them — if someone edits the Babel preset/target options without updating this constant, cached entries go stale silently. It also doesn't fold in the `@babel/preset-env` package version, so a dependency bump alone (loader file unchanged) would keep serving stale cached transpiles. Cheap fix: include `require('@babel/preset-env/package.json').version` in the key.

### Checked and held up (no issue found)

- `compress: false` restore for desktop/mobile (bug #2) — correct, verified against the Terser options.
- `check-build.yml` branch-filter restore (bug #3) — correct in isolation; just doesn't help *this* PR per finding #3 above.
- Filesystem cache: no cross-module (word/cell/slide/visio) collision — webpack namespaces its cache by the factory's own `name:` field per compiler.
- `.docker/sdkjs.bake.Dockerfile`: `SDK_PLATFORM=desktop` is correctly scoped to the parallel `sdkjs-desktop` stage only; doesn't leak into the web stage.
- `assertSafeBuildRoot` in `build/scripts/build-pipeline.cjs`: realpath-resolves symlinks, rejects the checkout/build dir and any ancestor containing the source tree — no bypass found.

</details>

MonaAghili:
@moodyjmz please review again.

moodyjmz:
## TL;DR

**Approve.** All five findings from my 23 July review are properly fixed — I verified each against the diff rather than taking the re-review request on trust, and CI is green on the exact head SHA (`c3abc4d121`, run `30011288062`, `event=pull_request`), which also independently demonstrates that finding #3 is resolved rather than merely claimed. The sentinel fix in particular is done the right way round: I re-derived webpack's stage constants from its own source to confirm the strip provably runs after Terser, and enumerated every consumer of `license.header` to confirm there's no third leak path.

My remaining reservations are **not about this PR's code**. They're merge sequencing: a certain file collision with #68, and the interaction with the 584-commit upstream sync sitting in #21. Details below.

<details>
<summary><strong>Verification of the five earlier findings</strong></summary>

**#1 sentinel leaking into every shipped file — fixed, and fixed correctly.**

`StripLicenseSentinelPlugin` (`build/webpack.sdk.factory.mjs:122`) strips post-minification; `build/scripts/deploy-assets.cjs:55` does the same for individually-deployed assets. Two things I checked rather than assumed:

- **Stage ordering holds.** Read from webpack 5.107.2's own `lib/Compilation.js`: `PROCESS_ASSETS_STAGE_ADDITIONS = -100`, `OPTIMIZE_SIZE = 400`, `DEV_TOOLING = 500`, `REPORT = 5000`. BannerPlugin injects at −100, Terser runs at 400, the strip runs at 5000. Correct by construction, and the inline comment explaining *why* it can't strip before injection is accurate.
- **No third consumer.** I enumerated all 30 build-directory scripts and grepped each for `license.header` / `license-banner`. Exactly two readers (`deploy-assets.cjs:46`, `factory.mjs:179`), both stripping. Nothing else can leak it.

**#2 bare-global fragility — addressed as suggested.** The WARNING block above `iife: false` names the three triggers (`import()`, `splitChunks`, externals) and states the failure mode. That's exactly what was needed.

**#3 CI not running on this PR — fixed and proven.** `main` added to both `push.branches` and `pull_request.branches`. All four checks now green on the head commit, so the `COMPILED=1` QUnit coverage added by bug-fix #4 has actually executed.

**#4 `build/DEVELOPER-GUIDE.md` — added** (261 lines). The benchmark table and the HMR-impossibility rationale are a genuinely useful addition beyond just satisfying the dangling references.

**#5 Babel cache key — fixed.** `presetEnvVersion: require('@babel/preset-env/package.json').version` folded into `BABEL_OPTIONS_KEY`, and the comment now describes what the code actually does rather than what it aspired to.

**New since my last pass, checked:** `c3abc4d121` sets `SDK_ADDONS=../../sdkjs-forms`. Path arithmetic is right — `npm run --prefix build` puts cwd at `sdkjs/build`, so `../../sdkjs-forms` resolves to the workspace-root sibling checkout.

**Looked like a bug, isn't — recording so nobody "fixes" it later.** The sentinel strip at stage 5000 runs *after* `SourceMapDevToolPlugin` at stage 500, so it mutates the `.js` after its map is computed. Harmless: it removes 19 characters from line 1 and no newline; line 1 is the injected banner comment, which carries no mappings; later lines don't shift. And since output filenames are fixed (`sdk-all.js`) rather than `[contenthash]`, the `realContentHash` pass at stage 2500 isn't affected either.

</details>

## Merge sequencing — two collisions

**1. `.eslintignore` is a certain add/add conflict with #68.** This PR creates it (16 lines: `deploy/`, `vendor/`, `pdf/build/`, `tests/`, …). #68 also creates it (one line: `common/Native/jquery_native.js`). Disjoint content, same new file — whichever lands second conflicts, and the reflexive "take theirs" resolution silently discards the other side's intent.

Worth knowing: **the `.eslintignore` added here is currently dead configuration.** Nothing in this PR runs eslint — `check-build.yml`'s `code-style` job is still only `python tests/code-style/check.py`. #68 is what actually invokes eslint. So the merged file wants the union of both lists, not either one alone.

**2. `check-build.yml` overlaps #68, probably benignly.** This PR doesn't touch the `code-style` job at all (only the `on:` triggers and the other two jobs), whereas #68 appends two steps to it. Different regions, so git should cope — though #68's insertion point sits immediately before `unit-tests:`, which this PR does modify, so the context lines may brush. Lower confidence than the `.eslintignore` collision, which is certain.

## The bigger sequencing question

```
main vs upstream master:      ahead_by=584  behind_by=76  status=diverged
#21 (update-from-master):     CONFLICTING / DIRTY, open since 2026-05-19
```

This PR deletes `build/Gruntfile.js` (561 lines, upstream-owned) and rewrites the build layer. Landing it *before* the upstream sync in #21 turns the build-layer portion of that already-conflicted 884-file PR from "hard" into "nobody will hand-resolve this."

I'd merge #21 first, then this. Not a criticism of the work — a consequence of the order.

Related: **#66 (`chore(deps): update dependency grunt to v1.6.3`) directly contradicts this PR.** It bumps grunt in `build/npm-shrinkwrap.json`; this PR deletes the Gruntfile and rewrites that lockfile. One of the two is waste regardless of the outcome. Suggest closing #66 if this lands.

## Caveats on this review

I did not run an independent second pass on this one. The conclusion I'd most want a cold reader on is my earlier finding #2 — the webpack single-module inlining behaviour that the bare-global model depends on. That rests on empirical build observation rather than plain source reading, and the WARNING comment added here is only as good as that premise.

Disclosure: #67 and #68 are mine, so read the sequencing notes with that in mind.

Nice work on the fixes — the sentinel one especially, since the obvious shortcut (strip before injection) would have quietly re-broken bug #1.

moodyjmz:
## Request: port web-apps' verify stage

Not a blocker, and not a criticism of the approach — the opposite. Now that I understand this is deliberately modelled on web-apps rather than a fresh toolchain choice, the pattern-match holds up well: per-product configs plus a shared factory (`webpack.{word,cell,slide,visio}.mjs` + `webpack.sdk.factory.mjs` ↔ `webpack.{documenteditor,spreadsheeteditor,…}.mjs` + `webpack.editor.factory.mjs`), and `build-pipeline.cjs` ↔ `build-pipeline.js`. Even the `mangle: false` decision matches, for the same reason — web-apps' factory documents it as *"117 source files use `var Common = Common || {}`"*, and sdkjs has the same bare-global exposure.

But the copy stops one phase short. web-apps has a verification tier this PR doesn't:

```
web-apps/build/scripts/
  verify-bundles.mjs
  verify-deploy.mjs
  verify-browser-target.mjs
  verify-replacements.mjs
```

`verify-replacements` runs as a `Preflight` phase before any build (`build-pipeline.js:246`, with the rationale "no point building if load-bearing idioms have drifted"), and the rest run as a post-build phase. `build-pipeline.cjs` here has the deploy-assets/webpack phase and the build-develop phase, then finishes.

**The specific ask: an sdkjs `verify-deploy` equivalent.** The highest-value assertion for this repo is that every path referenced by `DocumentServer/build/configs/core/DoctRenderer.config` actually exists in `deploy/` after a build. That config is what the converter reads at runtime — `core:DesktopEditor/doctrenderer/config.h:117-140` parses the `<file>` entries into `m_arrFiles`, and `editors.cpp:69-75` (`GetAllScript`) concatenates them ahead of `sdk-all.js`. A missing or renamed entry there produces a converter that fails at runtime with a green build, which is the exact failure class this repo just spent a week on (see Euro-Office/DocumentServer#307 — a converter-only defect in `common/Native/native.js` that no browser test could ever have caught).

Worth noting `Native/*.js` is in this PR's `deploy-assets.cjs` path (`'Native/*.js'`, and neither `native` nor `native_graphics` is in `IGNORE_NAMES`), so those files' handling changes here. A `verify-deploy` wouldn't validate their *contents*, but it would at least assert they're present and where `DoctRenderer.config` expects them.

Cheap version if a full port is too much scope for this PR: a single script that parses `DoctRenderer.config` for `<file>` entries, resolves each against `deploy/`, and exits non-zero on the first miss. Wire it as a step after `Run build sdkjs` in `check-build.yml`. Happy for that to be a follow-up PR rather than growing this one — say which you'd prefer.

---

Separately, and flagging it now rather than springing it later: there's a second gap I'd like to discuss, which is that nothing in this PR's CI exercises DoctRenderer or x2t at all — every check is browser QUnit via `node-qunit-puppeteer`, while `deploy-assets.cjs` adds a Babel `preset-env {targets:{ie:'11'}}` pass over files including the Emscripten glue that contains `async`/`await` (`common/zlib/engine/zlib.js`, `common/spell/spell/spell.js`, `common/hash/hash/engine.js`, `common/libfont/engine/fonts.js`, `pdf/src/engine/drawingfile.js`). That may well be like-for-like with what Closure's `--rewrite_polyfills=true` (`build/Gruntfile.js:460-463`) was already doing — I haven't established whether it's a new risk or a faithful port, and I don't want to raise it as an objection until I have. Mentioning it so it's not a surprise. `Euro-Office/sdkjs#34` (the e2e overlay job) looks like the natural gate for it.

moodyjmz:
## IE support is being dropped — two consequences for this PR

@MonaAghili — decision from our side: **we can drop IE support entirely.** That resolves a question left open in the earlier review and has two consequences here, one immediate and one worth a discussion.

### 1. The Babel ES5 pass can go — independent of anything else

`build/scripts/deploy-assets.cjs` runs `@babel/preset-env` with `{targets:{ie:'11'}}` over every non-ignored JS file. `IGNORE_NAMES` covers only the `_ie` variants, so the **non-`_ie`** files go through it — including the Emscripten glue that contains `async`/`await`: `common/zlib/engine/zlib.js`, `common/spell/spell/spell.js`, `common/hash/hash/engine.js`, `common/libfont/engine/fonts.js`, `pdf/src/engine/drawingfile.js`. `common/Native/native.js` and `native_graphics.js` are in that path too.

With no IE target, that transpile has no consumer. Removing it:

- deletes an `async`→regenerator downlevel that runs on the DoctRenderer path with no `@babel/runtime` present — the sharpest unverified risk in the diff, and the one I could not resolve in review
- removes `@babel/*` from the build dependency set
- drops the `BABEL_OPTIONS_KEY` cache-staleness surface in `loaders/sdk-concat.cjs` entirely

This is a subtraction, needs no toolchain change, and I would take it on its own merits.

### 2. With ES5 off the table, esbuild becomes viable

Three of the more intricate parts of this PR exist to make webpack behave like a concatenator: `output.iife: false` plus the single-module `sdk-concat-loader` (so bare top-level `var`s reach `window` — the ~287 `GlobalSkin`-style reads), `StripBootstrapStrictModePlugin`, and `BannerPlugin` + the `@@license-banner@@` sentinel + `StripLicenseSentinelPlugin`.

esbuild 0.25.12 does all three natively:

```
$ esbuild sdkjs-like.js --minify-whitespace --minify-syntax --banner:js='/* LICENSE BANNER */'
/* LICENSE BANNER */
var GlobalSkin={RulerDark:"#0f0f0f"},AscCommon={};window.AscCommon=AscCommon,...
```

True top level, no wrapper, identifiers untouched, banner prepended — no sentinel required. The three minify flags are separate (`--minify-whitespace` / `--minify-syntax` / `--minify-identifiers`), so `mangle: false` is a native mode rather than a Terser options object.

ES5 was the blocker and it is now gone:

```
✘ [ERROR] Transforming let to the configured target environment ("es5") is not supported yet
✘ [ERROR] Transforming async functions to the configured target environment ("es5") is not supported yet
```

Caveats, so this is a fair comparison rather than a pitch:

- My test was a synthetic file mimicking sdkjs's bare-global pattern, **not** a real 32 MB `sdk-all.js`. Indicative only.
- `--define:AscCommon.g_cIsBeta='"false"'` did **not** substitute — esbuild will not rewrite a member expression rooted at a symbol declared in the same file. Safer than DefinePlugin (same root cause as the LHS-assignment breakage your DEVELOPER-GUIDE documents), but defines would need doing as a text pass pre-concatenation. Upstream does exactly that: `build.py:146 apply_defines()`, a regex over `window.AscCommon.<name> = "..."`.
- I have no size or wall-clock numbers against your actual output.

**The parallelism win and the per-product-config + shared-factory pattern stand either way** — those are the strongest parts of this PR and I am not questioning them. The question is narrower: whether the concatenation half is worth ~5,500 lines of webpack machinery now that the ES5 constraint is gone.

Your call on whether that is worth exploring in this PR, a follow-up, or not at all. If the webpack route is still better on grounds I have not considered, I would rather hear that than have you re-open a settled decision.

### 3. Follow-up, needs care

The `_ie` variants (`zlib_ie`, `spell_ie`, `engine_ie`, `drawingfile_ie`, `fonts_ie`) become dead weight with IE dropped. `DoctRenderer.config` references none of them, so the converter path is unaffected — but something selects between `_ie` and non-`_ie` at runtime, so worth tracing before removal. Separate PR.

moodyjmz:
## Copyright headers on the new build files

Separate, small point — and to be clear up front, **this is inherited from web-apps rather than anything introduced here.** I've raised the equivalent issue against web-apps so it gets fixed in both places.

Every Euro-Office-authored file added by this PR carries:

```
 * (c) Copyright Ascensio System SIA 2010-2024
```

`build/lib/env.cjs`, `build/scripts/build-pipeline.cjs`, `build/scripts/deploy-assets.cjs`, `build/webpack.sdk.factory.mjs`, `build/loaders/sdk-concat.cjs`, `build/webpack.{word,cell,slide,visio}.mjs` and the rest. These are new files written in 2026; Ascensio System has no authorship claim on them, and the date range is stale even by ONLYOFFICE's own current headers (2009-2026 / 2012-2025).

**There's a CI catch that has to be fixed first.** `tests/code-style/check.py:37`:

```python
license_header = b'Copyright Ascensio System'
```

`check_file_without_license()` raises if any file lacks that exact string — and this PR *extends* that check to `.cjs`/`.mjs`. So right now a new Euro-Office file **cannot pass `code-style` without attributing itself to Ascensio System.** The header can't be corrected until the check is.

Suggested order:

1. Widen `check.py` to accept either an Ascensio header **or** an SPDX identifier. It can't simply be flipped — the ~400 genuine upstream source files legitimately carry Ascensio headers and should keep them. The check's purpose is "no unlicensed files", which either form satisfies.
2. Replace the headers on the files added here with an SPDX Euro-Office header, e.g.

```js
/**
 * SPDX-FileCopyrightText: 2026 Euro-Office contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

I couldn't find an established SPDX example in the Euro-Office tree to copy exactly, so treat that string as a proposal rather than the house form — worth confirming before applying it across files.

Note there's already a dual-attribution string in this PR at `build/scripts/deploy-assets.cjs:43`, used for generated output banners:

```
Copyright (C) Ascensio System SIA 2012-2025. All rights reserved; Euro-Office contributors 2026 - <year>
```

That's the right shape for *generated bundles*, which genuinely contain both parties' work. For new source files that contain only Euro-Office work, a plain Euro-Office header is the accurate one.

Happy for this to be a follow-up rather than growing this PR — but the `check.py` widening is a two-line change and it unblocks everything else.

MonaAghili:
@moodyjmz Can you please review again?

moodyjmz:
## TL;DR — Re-review

**Approve.** Every finding from the last round of review was fixed, and verified correct against source — not just taken on the commit message's word. One of the fixes (the bare-global regression test) goes further than what was asked. Two small residual items below, one of which I can now close out with hard evidence rather than leave open.

<details>
<summary><strong>Full re-review</strong></summary>

### Previously flagged, now fixed (verified against source)

**License-banner sentinel leak** — Fixed correctly in both places. `webpack.sdk.factory.mjs` adds a `StripLicenseSentinelPlugin` that strips `@@license-banner@@` from the asset *after* Terser's comment-matching pass has already used it to identify the real banner, so the match still works but the sentinel never ships. `deploy-assets.cjs` does the equivalent via `.replace(' @@license-banner@@', '')` on `licenseText` before use. Checked both replace paths — no stray artifacts left behind.

**Bare-global scoping fragility** — Handled better than asked. Not just a warning comment: `build/test/bare-global-contract.test.cjs` compiles a real minimal webpack config and proves *both* directions — the current shape (`iife:false`, single no-import entry) stays unwrapped and a bare `var` lands on the global object, **and** that adding `import()` (the documented danger case) actually does switch webpack to the wrapped runtime, so the test fails loudly rather than passing vacuously if that assumption ever breaks. webpack is also now pinned to an exact version (`5.108.4`, no caret) instead of a range — correct call, since this mechanism depends on an internal bundling heuristic that isn't part of webpack's public API contract and could silently change on a minor bump.

**CI never ran on this PR** — Fixed: `main` added to both `push` and `pull_request` triggers. Also went further: `tests/runAll.js` is now actually wired up as a real aggregate runner (previously dead code per AGENTS.md), and new `api-{word,cell,slide,visio}-compiled.html` suites load `sdk-all.js` directly rather than relying on `COMPILED=1`'s manifest, which never referenced `sdk-all.js` at all — that was the actual coverage gap. While wiring up the fuller suite, two pre-existing date-serial bugs were found and correctly identified as unrelated to this migration (confirmed via `git diff` against `main` showing no changes to the relevant formula code) and tracked as documented skips rather than hidden or used to block this PR.

**`build/DEVELOPER-GUIDE.md` dangling references** — Fixed; the file exists now (283 lines).

**Babel cache staleness** — Moot: the whole ES5-downlevel step was removed from the loader, so the stale-cache-key question no longer applies (see below).

### New from this round

**A. The sentinel-strip fix itself has no test coverage (low severity, easy fix).** `StripLicenseSentinelPlugin` isn't exported from `webpack.sdk.factory.mjs`, unlike its sibling `StripBootstrapStrictModePlugin`, which *is* exported and has both unit tests and a real-compilation integration test. Nothing in the test suite references `StripLicenseSentinelPlugin`. The fix for the bug that started this whole thread is correct today but unguarded — a future refactor could silently drop it from the plugins array and nothing would catch it. Suggest exporting the class and adding an integration test mirroring the existing `StripBootstrapStrictModePlugin` one.

**B. ES5 downlevel removal — resolved, not just documented as open.** A separate commit drops the entire Babel ES5-transpile pass (`sdk-concat.cjs` lost ~275 lines, `deploy-assets.cjs` lost its `transpileToES5`), so `let`/`const`/arrow functions/classes now ship as-authored on every platform including `SDK_PLATFORM=desktop|mobile`. `DEVELOPER-GUIDE.md` flags this honestly under an "Open Decision" heading and asks for sign-off on minimum runtime versions before closing it out — the right instinct, but I can actually close it:

- **Desktop:** `Euro-Office/desktop-apps` (`win-linux/defaults.pri`) pins `cef_version_107` — Chromium 107, October 2022. No other CEF version branch exists in that repo. Chromium 107 fully supports ES2022 (async/await, optional chaining, class fields, private methods) — nowhere close to being a concern for plain ES2015 syntax.
- **Mobile:** `web-apps/.claude/browser-floor-stage2-target.md` documents a deliberate, already-settled decision for the exact same question on Framework7/React hybrid mobile: **floor = iOS 17/Safari 17 → ES2022**. WKWebView on iOS 17 shares Safari 17's JavaScriptCore engine. Android is explicitly called out as *not* lowering the floor, since Android WebView is a Play-Store-updated component decoupled from OS version and tracks current Chrome regardless. Worth noting: that same doc traces a real mobile crash back to *over*-aggressive downleveling (an esbuild `target: 'es2015'` swap breaking i18next's class-based internals) — the fix there was to raise the target and stop transpiling, the same direction this PR is taking.

Recommend closing `DEVELOPER-GUIDE.md`'s "Open Decision" section with something like:

> **Resolved:** ES6+ is safe. Desktop pins CEF 107 (Chromium 107, ES2022-capable). Mobile floor is iOS 17/Safari 17 → ES2022 per `web-apps/.claude/browser-floor-stage2-target.md`, with Android WebView tracking current Chrome regardless of OS version. No downlevel step needed.

**C. Verified, not a problem: the one application-code change is legitimately justified.** `common/Drawings/Format/Controls.js` drops a duplicate `CControl`/`CControlPr`/`FormControlPr` factory registration — the PR body explains this is safe because `TableId.js`'s `private_InitFactoryClass()` re-registers the same three classes later anyway. I traced this independently rather than taking the claim at face value: `common/apiBase.js:3279` calls `AscCommon.g_oTableId.init(this)` at editor-construction time, which resets `m_oFactoryClass = {}` and repopulates it — confirming the removed early registration really was always getting clobbered. Confirmed dead code, fine as included.

</details>

moodyjmz:
## TL;DR

Ran an independent advisor pass over my own last two review comments before letting them stand — verified every load-bearing claim against the current head rather than trusting my own writeup. 7/8 held up exactly as stated; one (the `Controls.js` justification) was overstated in a way that doesn't change the conclusion but should be corrected rather than left wrong. Two small asks remain before merge, both optional/nice-to-have, not blockers.

<details>
<summary><strong>Detail</strong></summary>

### 1. Self-correction: finding C's justification was imprecise

My last comment justified keeping `common/Drawings/Format/Controls.js`'s dead-registration removal in scope by saying `TableId.js`'s `private_InitFactoryClass()` "re-registers the same three classes later anyway." That's not quite right — the `CControl`/`CControlPr`/`CFormControlPr` registration at `TableId.js:440-442` sits inside `if (window['AscCommonExcel'])`, so it does *not* fire unconditionally (e.g. not in a pure word/pdf context without the spreadsheet model loaded).

The actual reason the removal is safe: `CTableId.prototype.init()` (`TableId.js:51`) does `this.m_oFactoryClass = {}` *before* calling `private_InitFactoryClass`, unconditionally, regardless of that `AscCommonExcel` branch. `apiBase.js:3279` calls `g_oTableId.init(this)` at editor-construction time — after every chunk/module has already loaded. So whatever Controls.js registered early is wiped by that reset on every editor construction path, spreadsheet-model-present or not. `TableId.js` itself carries no diff in this PR (`git diff origin/main -- common/TableId.js` is empty), so post-init factory state is byte-identical to `main` either way. Conclusion unchanged — dead code, safe to remove — but the stated mechanism should be "the unconditional reset in `init()` clobbers it regardless" rather than "it gets re-registered anyway."

### 2. Still worth doing before merge (optional, not a blocker)

**Export and test `StripLicenseSentinelPlugin`.** It's declared as a bare `class` in `webpack.sdk.factory.mjs` (no `export`), unlike its sibling `StripBootstrapStrictModePlugin`, which is exported and covered by both a pure-function unit test and a real-compilation integration test. Nothing exercises `StripLicenseSentinelPlugin` today. It's the actual fix for the bug this whole review thread started over — right now it's correct but unguarded against a future refactor silently dropping it from the plugins array.

**Close out the "Open Decision" section in `build/DEVELOPER-GUIDE.md`.** It currently reads as unresolved ("needs sign-off... Not yet resolved"), but the sign-off already exists elsewhere in the org — it just isn't cited here. Suggested replacement text:

> **Resolved:** ES6+ is safe. Desktop pins CEF 107 (Chromium 107, Oct 2022 — ES2022-capable; see `Euro-Office/desktop-apps/win-linux/defaults.pri`). Mobile floor is iOS 17/Safari 17 → ES2022 per `web-apps/.claude/browser-floor-stage2-target.md`, with Android WebView tracking current Chrome regardless of OS version — that same doc traces a real mobile crash to *over*-aggressive ES2015 downleveling, the opposite direction of concern from this change. No downlevel step needed.

Neither of these gates the migration itself — both are cheap follow-ups, not reasons to hold the PR.

</details>

moodyjmz (approved):
Approving. Both follow-up items landed correctly and I verified each against source, not just the commit messages — `StripBundlePostprocessPlugin` is properly exported with a real-compilation test covering the sentinel strip, and the ES5 decision is closed out with a citation to a commit that's actually landed on `web-apps` main, not just a plan.

Good shout for going past the two asks, though: the dynamic `scanLimit` fix (the strict-mode scanner could have silently missed the directive once the banner grew past 2048 chars — real latent bug, predates this whole review thread), tracking `license.header` as a cache build-dependency (was invalidation gap, could've served a stale banner), and catching `SKIP_STANDALONE` missing from the `COMPILED=1` step (was silently clobbering the already-processed `device_scale.js`) are all things neither of my two review passes caught. That's the kind of thing that only turns up by actually re-deriving the mechanism instead of trusting it once it's green.

Also appreciated: going back and hardening `bare-global-contract.test.cjs` to compile the real `sdkConfig('word')` output instead of a synthetic stand-in, after noticing the synthetic version's loose `__webpack_require__` substring check would've false-alarmed against the real config's harmless `.g` helper. That's exactly the kind of "is my test actually testing the right thing" instinct that's easy to skip once a test is green.

Nice work.

PR #68, opened 2026-07-31
chore(lint): enable no-undef for the native browser shims

moodyjmz:
Follow-up to Euro-Office/sdkjs#67. **Merge #67 first** — see "Expected CI state" below.

## Why

`common/Native/` stubs `window`, `document`, `navigator`, `console`, `setTimeout` and `performance` for the embedded JS context that DoctRenderer/x2t runs. It is the one part of this tree that **no test ever executes**: `tests/runAll.js` and the `unit-tests` job drive QUnit in a real browser via `node-qunit-puppeteer`, where these stubs are never loaded and the real platform objects are used instead.

So an undeclared identifier in these files reaches production entirely unexercised. That is exactly how `console.error` came to reference an undeclared `param` and abort every converter operation that logged an error — silent save failure, ExitCode 86, data loss. Five weeks of dormancy, then production.

`no-undef` catches it. `.eslintrc.yaml` did not enable it, and nothing ran eslint at all — there is no eslint reference in `build/package.json`, no root `package.json`, and the `code-style` job only ran `tests/code-style/check.py` (license headers, LF endings, trailing newline).

## Proof it's a real guard

Against `main` as it stands today:

```
$ eslint --format unix 'common/Native/*.js'
common/Native/native.js:340:78: 'param' is not defined. [Error/no-undef]

1 problem
```

With #67's one-word fix applied: clean, exit 0. One error, the right error, no noise.

## Why it's scoped to common/Native

Repo-wide, `no-undef` reports **85,854 violations across 1,437 distinct identifiers**, measured per directory (common 26,446 / word 27,779 / cell 15,658 / slide 7,574 / pdf 6,543 / visio 1,854). The top offenders are legitimate cross-module namespaces:

| identifier | count |
|---|---|
| `AscCommon` | 16,525 |
| `AscDFH` | 16,021 |
| `Asc` | 13,402 |
| `AscFormat` | 12,797 |
| `AscCommonExcel` | 4,712 |

Making that green would need a globals list larger than the rule is worth, so this PR does not attempt it. Widening the scope later is a separate conversation with a separate cost.

## Why `browser: false`

Counter-intuitive but load-bearing. These files *provide* the browser globals rather than consuming them. With `browser: true` the existing `no-redeclare` rule fires on every stub — `navigator`, `document`, `console`, `setTimeout`, `setInterval`, `clearTimeout`, `clearInterval`, `performance`, `Image` — and buries the one real defect in eleven false ones. I hit exactly that while building this; the first draft reported 12 problems, of which 11 were noise.

With `browser: false`, the three globals the converter genuinely injects before any script runs are declared explicitly, each pointing at where `core` installs it:

| global | installed by |
|---|---|
| `window` (the global object itself) | `doctrenderer.cpp:579` — `global_js->set("window", global_js)` |
| `native` | `doctrenderer.cpp:583` — `global_js->set("native", oNativeCtrl)` |
| `CreateEmbedObject` | `js_internal/v8/v8_base.cpp:218` — `InsertToGlobal(...)` |

Split into two per-file override blocks because `native.js` declares `Asc`/`AscCommon`/`AscFormat`/`AscDFH` itself, whereas `native_graphics.js` consumes `AscCommon` from it — `editors.cpp` (`GetAllScript`) concatenates them in that order. Declaring `AscCommon` globally for both would trip `no-redeclare` on `native.js`'s own `var AscCommon = {}`.

`jquery_native.js` is ignored via `.eslintignore`, matching `tests/code-style/check.py`'s own `exclude_files`.

## Expected CI state

**`code-style` will fail on this branch until #67 merges**, with exactly:

```
common/Native/native.js:340:78: 'param' is not defined. [Error/no-undef]
```

That failure is the demonstration, not a defect in this PR. Merge #67, and this goes green with no changes.

## Divergence cost

`.eslintrc.yaml` and `.github/workflows/check-build.yml` both exist upstream, so this is real fork divergence. It is confined to an `overrides:` block appended to the end of one file and one step added to one job — upstream touches neither often — and `.eslintignore` is a new file, so it cannot conflict. Judged worth it: the alternative is relying on nobody ever again writing an undeclared identifier in the only directory CI cannot reach.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+# Vendored jQuery fork. tests/code-style/check.py excludes it as well.
+common/Native/jquery_native.js
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -36,3 +36,49 @@ rules:
   # Stylistic Issues
   # http://eslint.org/docs/rules/#stylistic-issues
   linebreak-style: [error, unix]
+
+overrides:
+  # The files in common/Native stub the browser environment for the embedded JS
+  # context used by DoctRenderer/x2t. Nothing in CI executes that bundle - the
+  # QUnit suites run in a real browser, where these stubs are never loaded - so
+  # an undeclared identifier here reaches production completely unexercised.
+  # That is how console.error() came to reference an undeclared `param` and
+  # abort every converter operation that logged an error.
+  #
+  # no-undef is enabled for this directory only. Enabling it repo-wide would
+  # report ~86k violations across 1437 identifiers, almost all of them
+  # legitimate cross-module namespaces (AscCommon, AscDFH, Asc, AscFormat, ...),
+  # which would need a globals list larger than the rule is worth.
+  # `browser: false` in both blocks below is deliberate and load-bearing. These
+  # files *provide* window/document/navigator/console/setTimeout/performance for
+  # a context that has none. Treating them as browser code makes every stub read
+  # as a redeclaration of a built-in and buries the real defects in the noise.
+  #
+  # The two host globals are injected by the converter before any script runs:
+  #   window             -> the global object itself, doctrenderer.cpp:579
+  #   native             -> doctrenderer.cpp:583 (global_js->set("native", ...))
+  #   CreateEmbedObject  -> js_internal/v8/v8_base.cpp:218 (InsertToGlobal)
+  - files:
+      - common/Native/native.js
+    env:
+      browser: false
+      es2015: true
+    globals:
+      window: readonly
+      native: readonly
+    rules:
+      no-undef: error
+
+  - files:
+      - common/Native/native_graphics.js
+    env:
+      browser: false
+      es2015: true
+    globals:
+      window: readonly
+      CreateEmbedObject: readonly
+      # Declared by native.js, which editors.cpp (GetAllScript) concatenates
+      # ahead of this file.
+      AscCommon: readonly
+    rules:
+      no-undef: error
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -22,6 +22,18 @@ jobs:
       - name: execute check styles
         run: python tests/code-style/check.py
 
+      - name: setup node
+        uses: actions/setup-node@v6
+        with:
+          node-version: 20
+
+      # Scoped to common/Native on purpose - see the overrides block in
+      # .eslintrc.yaml. Those shims are the one part of the tree that no test
+      # ever executes, so no-undef is the only thing standing between an
+      # undeclared identifier and a converter that aborts on every logged error.
+      - name: lint native shims
+        run: npx --yes eslint@8.57.1 'common/Native/*.js'
+
   unit-tests:
     runs-on: ubuntu-latest
     steps:


moodyjmz:
CI has now run, and the guard behaves exactly as intended. `code-style` fails with a single error and nothing incidental:

```
340:78  error  'param' is not defined  no-undef
✖ 1 problem (1 error, 0 warnings)
Process completed with exit code 1
```

`DCO`, `build-test` and `unit-tests` all pass. The `code-style` failure is `main`'s existing defect being caught, not a problem with this PR — it clears once #67 merges. Steps 1–5 of the job (checkout, python, `check.py`, node setup) all passed, so eslint genuinely ran and rejected the code rather than the step falling over.

## Two caveats in the `lint native shims` step that reviewers should weigh

Both are visible in the CI log but not stated anywhere in the PR, which is a gap on my part.

**1. eslint 8.57.1 is end-of-life.** The log says so on every run:

```
npm warn deprecated eslint@8.57.1: This version is no longer supported
```

I pinned 8.x because eslint 9 requires flat config (`eslint.config.js`), and `.eslintrc.yaml` is an upstream-owned file. Migrating it would convert a 60-line change into a config migration with a much larger divergence footprint and a guaranteed conflict on the next ONLYOFFICE sync. That trade still looks right to me, but the consequence is that this step is pinned to a dead-end major version and the warning will nag indefinitely.

**2. `npx --yes` fetches eslint from the registry on every CI run.** No lockfile, no integrity pin, no cache. That means:

- a supply-chain surface — the step resolves and executes a freshly-downloaded dependency tree each run, with only the version string pinned
- a build-reliability dependency on npm being reachable, in a job that previously needed nothing but Python
- roughly a second of install time per run, which is noise, but it is not zero

## The cleaner alternative, and why I didn't just do it

A root `package.json` with eslint as a pinned devDependency, `npm ci`, and `actions/setup-node`'s `cache: npm`. That gets a lockfile, integrity hashes, and a warm cache — strictly better on all three counts above.

I did not do it because **sdkjs has no root `package.json` at all** — only `build/package.json`. Introducing one to a mirrored fork is a structural change with consequences beyond this PR (it changes what tooling infers about the repo root, and it is a new file upstream may later add itself), and that is a call for whoever owns the repo's build surface rather than something to slip into a lint fix.

So: the scoped rule and the config are what I am actually asking to be reviewed here. If the lockfile is worth a root `package.json`, say so and I will do it as a follow-up rather than expanding this PR.

moodyjmz:
## Merge Euro-Office/sdkjs#65 before this PR

#65 (the Grunt → webpack build migration) touches two of the same files, and one of them is a guaranteed conflict. It's the larger, older PR and it's green — this one should rebase onto it, not the other way round.

**1. `.eslintignore` — certain add/add conflict.** Both PRs *create* the file, with disjoint content:

| | contents |
|---|---|
| **#65** | `deploy/`, `develop/`, `build/node_modules/`, `vendor/`, `common/zlib/`, `common/libfont/`, `pdf/src/engine/drawingfile_ie.js`, `pdf/build/`, `tests/` |
| **this PR** | `common/Native/jquery_native.js` |

Whichever lands second conflicts, and the reflexive "take theirs" resolution silently discards the other side's intent. **The resolution must be the union of both lists** — #65's entries are the useful baseline for any future repo-wide lint, and this PR's single entry is what the scoped `common/Native` rule needs in order to skip the vendored jQuery fork.

Worth noting the two are complementary rather than redundant: #65 adds an `.eslintignore` but never actually runs eslint anywhere — its `code-style` job is still only `python tests/code-style/check.py`. This PR is what wires eslint in. So today #65's ignore file is inert and this one is load-bearing; after both land, they're one list serving one invocation.

**2. `check-build.yml` — overlapping but probably benign.** #65 doesn't modify the `code-style` job at all (it changes the `on:` triggers and the other two jobs), while this PR appends two steps to `code-style`. Different regions, so git should cope. Flagging it because this PR's insertion point sits immediately before `unit-tests:`, which #65 *does* modify, so the context lines may brush. Lower confidence than the `.eslintignore` collision, which is certain.

## Correction to this PR's description

The body states that `check-build.yml` has "a bare `pull_request:` trigger with no branch filter." That is true of `main` today and **will be false once #65 merges** — #65 restores the `pull_request.branches` filter and adds `main` to it (commit `de2b1dd41e`).

No functional impact on this PR: `main` is in the restored filter, so PRs targeting `main` still get CI. But the reasoning in the description goes stale, so don't read it as current after #65 lands.

## Suggested order

1. **#67** — independent of all of this. It only touches `common/Native/native.js`, which #65 does not modify at all, and it's the one-line fix for the actual data-loss bug. No reason to hold it behind a 5,493-line build migration.
2. **#65** — larger, older, green, and it should precede the upstream sync discussion in #21.
3. **This PR** — rebase onto #65, merging `.eslintignore` to the union above and dropping the stale trigger claim from the description.

MonaAghili:
**Confirmed accurate:**
- Ran `eslint@8.57.1` against `common/Native/*.js` with this PR's config: exactly one violation, `native.js:340:78 'param' is not defined` — matches the "proof it's a real guard" section exactly.
- `native_graphics.js` lints clean under its override; the `browser: false` / `no-redeclare` reasoning is correct (`native.js` does declare `navigator`/`window.*` via `var`, which `browser: true` would flag).
- The `doctrenderer.cpp:579`/`:583` and `v8_base.cpp:218` citations are accurate to the line.
- `jquery_native.js` exclusion matches `check.py`'s own `exclude_files`.

**On the two caveats you raised yourself (EOL eslint pin, `npx --yes` supply-chain surface):** agreed on both, and agreed a root `package.json` is out of scope for this PR specifically — that's a repo-structure call, not a lint fix, and shouldn't ride in here.

**One thing that needs a decision before merge order is final:** the "merge #65 first, guaranteed `.eslintignore` conflict" comment doesn't match #65's diff as of its latest commit (2026-08-06, six days after that comment). #65 currently touches `.gitignore` and `check.py`'s `exclude_dirs` — I don't see it creating or touching `.eslintignore` at all right now, so the union-merge conflict described may already be resolved or may have been describing an earlier version of #65. Can you confirm whether #65 still creates `.eslintignore` on your end? If it doesn't anymore, the merge-order note in your comment is stale and worth striking so nobody spends time resolving a conflict that isn't there.

Everything else — the `no-undef` scoping rationale, the `browser: false` design, the `native.js`-before-`native_graphics.js` load-order dependency (which lives entirely in DocumentServer's `config->m_arrFiles`, worth a one-line note in AGENTS.md at some point but not a blocker) — holds up. Approving pending #67, with the #65 status above cleared up first.

<details>
<summary>Full verification log and severity-tagged findings</summary>

**Scope reviewed:** `gh pr diff 68` (60 additions / 0 deletions, 3 files: `.eslintignore`, `.eslintrc.yaml`, `.github/workflows/check-build.yml`) — no runtime SDKJS code changes. Verified every factual claim in the PR description against the live sdkjs checkout and the DocumentServer/core repo (`doctrenderer.cpp`, `v8_base.cpp`, `editors.cpp`).

### Verification performed
- Ran `npx eslint@8.57.1 --format unix 'common/Native/*.js'` with this PR's `.eslintrc.yaml`: exactly one violation, `native.js:340:78 'param' is not defined`, matching the PR description exactly.
- Confirmed `native_graphics.js` lints clean (0 problems) under its own override block — the `AscCommon`/`window`/`CreateEmbedObject` globals list is complete today.
- Confirmed `native.js` declares `navigator`, `document`-adjacent globals, and `window.*` via `var` (`common/Native/native.js:27`) — exactly what would collide with `no-redeclare` under `browser: true`.
- Cross-referenced `doctrenderer.cpp:579` and `:583` — `global_js->set("window", global_js)` and `global_js->set("native", oNativeCtrl)` are exactly there. `v8_base.cpp:218` confirms `InsertToGlobal("CreateEmbedObject", ...)`.
- Confirmed `tests/code-style/check.py` excludes `jquery_native.js` via `exclude_files = {'jquery_native.js'}`, matching the new `.eslintignore`.
- Confirmed no root `package.json` and no `eslint` reference in `build/package.json` — the "nothing currently runs eslint" claim holds.
- Could **not** fully verify that `editors.cpp`'s `GetAllScript` concatenates `native.js` before `native_graphics.js`. `native_graphics.js` is bundled into `sdk-all.js` via `configs/word.json:457`, but `native.js` itself doesn't appear in any `configs/*.json` — it's copied standalone by the Gruntfile's `copy` task and must be prepended via `config->m_arrFiles` on the DocumentServer side, outside this repo. Plausible, unconfirmed.

### Findings

🟢 **Low — `native.js`/`native_graphics.js` load-order dependency is enforced entirely outside this repo, with nothing pinning it.** The `native_graphics.js` override's `AscCommon: readonly` depends on `native.js` running first. That ordering lives in DocumentServer's `config->m_arrFiles`, not sdkjs, and nothing asserts it. Recommendation: a one-line note in AGENTS.md/CLAUDE.md; not a merge blocker.

🟢 **Low — pinned eslint version (`8.57.1`) is already EOL upstream.** Already acknowledged by the author above; agreed non-blocking.

💡 **Suggestion — `es2015: true` in both override blocks is broader than these files need.** Pulls in `Promise`/`Symbol`/`Map`/`Set` globals not obviously used by these two small stub files. Harmless; `es5: true` would be tighter but not worth a change request.

💡 **Suggestion — CI failure mode is intentional but could be misread.** This branch's `code-style` job fails on its own until #67 merges by design. Risk only if someone force-pushes or an automated process treats red CI as broken rather than "merge order matters."

### Final Assessment

**Overall Review:** Ready to Merge, sequenced after #67.
**Risk Score:** 2/10 — pure CI/tooling addition, zero runtime code changes, blast radius limited to "a CI step can go red."
**Biggest Strengths:** every claim in the PR description independently verified; correctly scoped vs. the ~86k-violation repo-wide alternative; `browser: false` justified with evidence, not assertion; minimal fork-divergence cost.
**Biggest Risks:** the cross-repo load-order assumption has no test pinning it; the guard only catches undeclared-identifier bugs, not logic bugs, in `common/Native/`.
**Missing Tests:** `common/Native/*.js` still has zero executable test coverage (QUnit runs in a real browser and never loads these stubs) — worth a follow-up ticket for a Node-based smoke test, independent of this PR.

</details>

PR #70, opened 2026-08-03
SpreadSheet Editor Dark Mode (dark document) 🌘 — companion to web-apps PR

j-base64:
# Origin

Issue https://github.com/Euro-Office/web-apps/issues/117 pinpoints the lack of dark mode in the spreadsheet editor, unlike the document editor, which already has one.

Originally noticed on web-apps, the fix requires changes to both `sdkjs` and `web-apps`, plus optionally `DocumentServer` for e2e smoke tests.

# Approach

This PR brings dark mode (dark document feature) to the spreadsheet editor with a simple starting approach, rather than full negative-inversion or full token-based re-theming:

https://github.com/user-attachments/assets/15f06c8f-5b87-4d12-bf7e-31683f990629

## What this dark mode DOES ✅
- **Define & Apply a dark mode colorset to background, text, and border color of cells left on automatic colors¹**
- **Honors print/print preview with an always light theme**
- **Sets automatic text to white on cells with an explicit dark background** (see Design Notes)
- **Handles**:
  - **Partially colored text**
  - **Merged and conditionally-formatted cells**
  - **Search-highlighted cells**
  - **Row/column guide color when resizing**

## Design notes 🌈
- To maximise compatibility of designed cells in light mode, the automatic text color of cells with an explicit background is computed the same way as in light mode. This led to a design choice: flip automatic text from black to white on darkened cells, so a user isn't surprised to set a black background on a cell and find its text invisible. However, users can override text color with an explicit color. Try it and see.
- In the future, it probably makes sense to simplify by merging the dark theme and dark document features into one sun/moon-style switch, like many modern products do.

## What this dark mode DOESN'T DO ❌
- Affect inserted elements like images, shapes, charts
- Fix certain table templates that use transparent instead of white to draw "white" rows
- Rely on inverting colors of the document (negative method) or the UI color palette
- Ship on mobile

## Known tradeoffs 
- **Equations** are not supported yet (remains black), should be fixed on an update
- A cell being actively edited keeps its previous theme's colors until you click away. Prototyped a fix, reverted it, not worth the complexity for a small, self-correcting gap.

_¹ Automatic color = different from explicit color (those explicitly set by the user or a table style template)_

# Basic manual testing
1. Create content and colorisations on a sheet
   - Type text in a few cells, set an explicit/custom text color on a whole cell, and on just some words or letters within another cell.
   - Add a border to two cells, set explicit/custom border color for one of the cells.
   - Add a predesigned table 5 cols x 5 rows.
2. Under `View` tab > Make sure the interface theme is Dark.
3. Switch `Dark Document` button to see differences.
4. In dark mode
   - Search for some text.
   - Resize a column or row. The drag guide should be visible against the dark background.
   - Open Print Preview.

# Automated testing
- Proposed on `DocumentServer` see below, 13 smoke tests covering cell fill/border/text correction, merged cells, page-break-preview, gradient/pattern fill in the cell editor, print-preview, column-resize guide color, search-highlight contrast, mid-edit-toggle behavior, interface-theme interaction, and toolbar sync

# Code

**This PR spans BOTH `sdkjs`, `web-apps`,** and optionally `DocumentServer`:

- **YOU'LL NEED additionaly**
   - This `web-apps` PR for the UI part: https://github.com/Euro-Office/web-apps/pull/187.
- Optionally: 
   - This `DocumentServer` companion PR for e2e/smoketests: https://github.com/Euro-Office/DocumentServer/pull/310.

## AI assistance

- Developed with AI assistance (Claude Code, mainly `claude-sonnet-5`)
  
```
🌘 First PRs in the project, feedback and reviews welcome :)
```


MonaAghili:
# PR #70 Review — SpreadSheet Editor Dark Mode

**Scope reviewed:** `gh pr diff 70` (362 additions / 38 deletions, 10 files) plus surrounding code read directly from the checkout to verify call-site completeness (all `drawFillCell` callers, all `_drawText`/`render` callers, `isColorAutomatic`, print-path branching, `WorkbookView` construction order).

---

## 1. Correctness

🟡 **Medium — `drawFillCell`'s dark-mode correction is unconditional per-call, not gated on "is this color automatic"**
[cell/utils/utils.js:2538](cell/utils/utils.js#L2538) (`drawFillCell`)
Unlike the text/border paths (which all gate on `AscCommonExcel.isColorAutomatic(color)` before correcting), `drawFillCell` corrects *any* solid fill whenever `ctx.isDarkMode` is true, unless the caller explicitly passes `bKeepsFillColorAsIs=true`. Three pre-existing callers were not updated to pass this flag: [cell/utils/utils.js:2439](cell/utils/utils.js#L2439), [:2687](cell/utils/utils.js#L2687), [:2850](cell/utils/utils.js#L2850) — these render cell/table **style-gallery thumbnails** using the style's literal, explicit fill color.
**Why it matters:** today this is inert because thumbnail `DrawingContext` instances default `isDarkMode = false` and are never toggled. But the correctness invariant ("only automatic colors get corrected") is enforced by caller discipline rather than by the function itself. The very next natural feature request — "preview style swatches against the active theme" — would silently corrupt these three call sites' literal colors with no compile-time or review signal, because the guard lives in the wrong place.
**Fix:** either gate the correction inside `drawFillCell` on `isColorAutomatic(solid)` (matching the border/text pattern), or invert the default so correction is opt-in (`bApplyDarkModeCorrection`) rather than opt-out — a missed opt-in fails safe, a missed opt-out doesn't.

🟢 **Low — mixed print-context signaling (`isPrint` param vs `this.usePrintScale` instance flag)**
[cell/view/WorksheetView.js:6097](cell/view/WorksheetView.js#L6097) computes `isPrint = this.usePrintScale` and reuses it locally, while the sibling branch at [:6182](cell/view/WorksheetView.js#L6182) re-reads `this.usePrintScale` directly instead of the already-computed `isPrint`. Functionally identical today (verified: every `printState` vs `defaultState` branch point does correctly special-case print), but it's two idioms for one concept in the same function, which invites drift the next time someone edits one branch and not the other.

💡 **Suggestion — leftover AI-translation comment markers shipped in code**
[cell/view/WorksheetView.js:6116](cell/view/WorksheetView.js#L6116) and [:6115](cell/view/WorksheetView.js#L6115) contain literal `/** ↪ AI Translation → "..." */` comments (artifacts of translating pre-existing Russian comments). Harmless, but reads as unpolished for shipped code — reword as plain English comments before merge.

---

## 2. Regression Analysis

I specifically checked for missed spots where the pre-existing "always light" print/print-preview guarantee could regress under dark mode, since that's the highest-blast-radius regression class here (wrong-looking printed/exported output is very visible to end users). Traced every remaining `cells.defaultState.background/border` call site in `WorksheetView.js` (frozen-pane scroll-shift redraw, `_clean()`, pivot-collapse icon, style thumbnails) — all are screen-only paths untouched by print rendering, and every genuine print/print-preview/page-break path was correctly repointed to `cells.printState`. **No regression found here** — this is a real strength of the implementation (see §11).

I also checked whether `isColorAutomatic`'s identity check for font color survives cloning through the undo/redo and copy/paste paths (`RgbColor.clone()`, `ThemeColor.clone()`) — confirmed `ThemeColor.clone()` returns `this` (so identity is preserved) and `RgbColor.isAutoColor` is copied explicitly in the new `clone()` override. No regression found in the color-identity plumbing itself, but see the architecture concern below — the *mechanism* is fragile even though it's currently correct.

No regressions found in serialization: `WriteColorSpreadsheet` and friends write explicit fields, not a generic object dump, so the new `RgbColor.isAutoColor` runtime flag cannot leak into saved `.xlsx`/binary format and cannot cause a save/open round-trip corruption.

---

## 3. Architecture

🟠 **High — two incompatible mechanisms for "is this color automatic," one flag-based, one identity-based**
[cell/model/WorkbookElems.js:148](cell/model/WorkbookElems.js#L148) (`isColorAutomatic`)
`RgbColor` gets an explicit `isAutoColor` boolean that survives cloning. `ThemeColor` (used for the workbook's default font color) has no such flag — instead the function relies on `color === g_oDefaultFormat.Font.c`, i.e., object identity against a mutable module-level singleton, which only works because `ThemeColor.clone()` happens to return `this` rather than a real copy.
**Why it matters:** this is a hidden coupling between three unrelated pieces of code (`isColorAutomatic`, `ThemeColor.prototype.clone`, and wherever `g_oDefaultFormat.Font.c` gets assigned in `Serialize.js`) with no assertion anywhere that the invariant holds. A future contributor "fixing" `ThemeColor.clone()` to do an actual deep clone (a very plausible, reasonable-looking change) would silently break automatic-text detection for the default font color — text would either always or never get dark-mode-corrected, with no test to catch it, because there is no test (see §9).
**Recommendation:** give `ThemeColor` the same explicit `isAutoColor`-style flag as `RgbColor` rather than leaning on identity+clone semantics. It's a few more lines but removes the invisible cross-file contract.

🟢 **Low — `for (var i in this.wsViews)` in `WorkbookView.prototype.updateDarkMode`**
[cell/view/WorkbookView.js:5352](cell/view/WorkbookView.js#L5352) iterates an array with `for...in` immediately after using `.forEach` on `_getOwnedDrawingContexts()` in the same method — inconsistent style for no reason, and `for...in` on arrays is a long-standing footgun (picks up enumerable prototype properties). Trivial fix: `for (var i = 0; i < this.wsViews.length; i++)` or `.forEach`.

💡 **Suggestion — scope creep in `common/skin.js`**
The `"" === theme[colorMap[color]]` fix in `updateGlobalSkinColors` ([common/skin.js:760](common/skin.js#L760)) is a legitimate, well-justified fix (documented inline: web-apps' CSS-custom-property-backed theme objects report missing values as `""`, not `undefined`), but it's an interface-theme (`GlobalSkin`) bug fix bundled into a content-dark-mode PR. Not wrong, just worth calling out explicitly in the PR description/changelog since it changes behavior for the *existing* interface-theming system, not just the new feature.

---

## 4. Performance

No O(n²) or per-frame-allocation concerns found:
- `DrawingContext._darkModeRgbCache` and the shared `_darkModeColorShuttle` avoid per-pixel/per-call allocation — the shuttle-object pattern is a legitimate, documented technique here (every consumer reads it synchronously), not a latent aliasing bug.
- `_getKeepsAutomaticTextColorAsIs`'s `this.workbook.inFindResults(this, row, col)` call (line 6498) duplicates an *existing* expression already used identically at line 6137 for the same purpose — so the per-cell cost is not new, it's the same cost the codebase already pays for search-highlight lookup during every redraw, just invoked from a second call site now. Worth extracting into a small shared helper (`_getFindHighlightColor(row, col)`) to avoid the literal duplication, but not a performance regression.
- Cache growth in `_darkModeRgbCache` is bounded by the number of distinct RGB values actually used in a document — negligible for real spreadsheets.

Scales fine for large documents; no new per-cell deep clones or DOM writes introduced.

---

## 5. JS Quality

- The shared mutable `CColor` "shuttle" object returned by `getDarkModeCorrectedColor` is a sharp-edged pattern (a return value that a caller must not retain) but is well-documented via JSDoc, and I verified every current call site consumes it synchronously before the next call — no aliasing bug today. Worth a comment at the definition site warning against future async use, since the risk is invisible at the call site itself.
- `isColorAutomatic` and the new dark-mode branches are consistently null-guarded (`!color` returns `true`, i.e., "no color = automatic" — correct default).

---

## 6. Maintainability

- The `_getKeepsAutomaticTextColorAsIs(c, row, col, resolvedFallbackBg)` function does three different jobs behind one boolean return (search-highlight check, solid-fill contrast check, pattern/gradient fallback) and is called from two different draw paths with different arguments (grid vs. cell editor). The logic is correct and well-commented, but a future reader modifying the grid path is likely to miss that the cell-editor path has a materially different code path (the `resolvedFallbackBg` branch) for the same-named concept. Consider splitting into `_isSearchHighlighted`, `_isFillLightEnough` as named building blocks the two call sites compose explicitly, rather than one function branching on an optional 4th argument.

---

## 7. Build/Webpack
No build/webpack config in this diff — N/A.

---

## 8. Security
No user-controlled input, no HTML generation, no eval — colors are numeric RGB manipulation only. No findings.

---

## 9. Testing

🟠 **High — zero automated tests land in this repo with this PR**
`cell/` has no existing unit-test directory in this repo (only `pdf/test`, `common/geometry/test`, `common/libfont/test` exist), and this PR adds none. The 13 proposed smoke tests live entirely in the separate `DocumentServer` companion PR (#310) as e2e/visual tests, which:
- won't run as part of this repo's own CI/PR checks,
- can't isolate regressions to the specific unit (`isColorAutomatic`, `getDarkModeCorrectedColor`, `_getKeepsAutomaticTextColorAsIs`) when a screenshot diff fails.

**Missing test scenarios** (unit-level, independent of the e2e smoke suite):
- `isColorAutomatic`: RgbColor auto vs. explicit-black(0) vs. cloned-auto; ThemeColor default-font-color identity survives a full charProperties clone chain.
- `getDarkModeCorrectedColor`: cache hit/miss returns the same RGB for repeated calls; shuttle object is correctly overwritten (not stale) across consecutive different-color calls.
- `_getKeepsAutomaticTextColorAsIs`: solid light fill → true; solid dark fill → false; pattern/gradient with no `resolvedFallbackBg` → true (conservative); pattern/gradient with `resolvedFallbackBg` → resolves against that instead; search-highlighted cell overrides fill-based result.
- Print/print-preview: rendering with `isDarkMode=true` + `usePrintScale=true` produces `printState` colors, not `defaultState`.
- Regression guard for the exact bug class already fixed once in this branch's own history (`591ca3151e`/`42705a02f4`/`12c3905cdf`: data-bar colors, gradient/pattern editor text, theme corruption) — these were manually caught and fixed during development; none of them has a regression test pinning the fix, so they can silently reappear.

---

## 10. Code Style
Nothing beyond what's covered above (AI-translation comment artifacts, `for...in` on an array). Not flagging further — the new code is otherwise clean, consistently commented, and follows existing naming conventions in the file.

---

## Final Assessment

**1. Overall Review: Merge with Minor Changes**
The core algorithm (automatic-vs-explicit color detection, print/print-preview isolation, search-highlight and merged-cell handling) is correct, and I could not find an actual regression by tracing every touched draw path. The findings above are real but narrow: one latent robustness gap (`drawFillCell` guard placement), one architectural fragility (`ThemeColor` identity-based auto-detection) that's correct today but has no safety net, and a testing gap that's more about this repo's norms than this PR's diligence — the branch's own commit history shows the author already went through several rounds of self-caught dark-mode color bugs.

**2. Risk Score: 4/10**
Contributing factors: an isolated, additive feature (no existing rendering path is removed, only conditionally branched on `isDarkMode`, which is false by default and requires an explicit user toggle); no document-model/serialization changes; no collaborative-editing surface touched. Pulled up from a lower score by: zero regression tests landing with the change, and one hidden cross-file invariant (`ThemeColor.clone()` returning `this`) that nothing asserts.

**3. Biggest Strengths**
- Careful, correctly-scoped separation of "interface theme" (`GlobalSkin`) from "content dark mode" (`EditorSkins` + `isDarkMode`) — verified these are genuinely independent axes, matching the PR's stated design goal (light UI + dark document, and vice versa).
- Print/print-preview isolation (`printState` vs `defaultState`) is thorough — I checked every remaining background/border call site in `WorksheetView.js` and found no leak of dark colors into printed output.
- Good judgment on scope: explicit color detection (`isColorAutomatic`) correctly preserves user- and template-authored colors untouched, only correcting genuinely automatic ones — this is the hard part of "dark mode without full re-theming" and it's implemented soundly.
- Visible evidence of iterative self-review (branch history shows three targeted bug-fix commits for data bars, gradient/pattern editor text, and theme corruption) — this is exactly the kind of edge-case hunting a reviewer wants to see already done.

**4. Biggest Risks**
- No regression tests for a rendering feature that has already needed three follow-up bug fixes during development — any future refactor of `StringRender`, `WorkbookView`, or the color-model classes has no automated tripwire.
- The `ThemeColor.clone() === this` identity contract is load-bearing and undocumented at its point of truth (only commented at the *consumer* in `WorkbookElems.js`, not at `ThemeColor.prototype.clone` itself).
- `drawFillCell`'s opt-out (rather than opt-in, or self-checking) correction default is a footgun for the next feature that touches thumbnail/preview rendering.

**5. Missing Tests**
See §9 in full — summarized: unit coverage for `isColorAutomatic`, `getDarkModeCorrectedColor` caching, `_getKeepsAutomaticTextColorAsIs` contrast branches, and print-mode color resolution; regression pins for the three bugs already fixed in this branch's history.

**6. Final Recommendation**
I would approve this with the `drawFillCell` guard-placement issue addressed (either gate on `isColorAutomatic` inside the function or invert the flag's default) and a request to land at least the unit-test-shaped scenarios above in `cell/` before or shortly after merge — not because I found a live bug, but because a render-path feature that's already needed three self-caught fixes during development is exactly the kind of change that benefits most from a regression net, and this repo currently has none for `cell/`. The architecture concern (`ThemeColor` identity coupling) is worth a follow-up ticket rather than a merge-blocker, given it's correct today and documented at the consumer site.

j-base64:
Thanks a lot for the review and spoting these points 🙌

I followed your final recommendation improving required aspects and did my best on the others improving them or keeping them as follow-ups.


## Required

- ✅ **Done**
  - **drawFillCell guard-placement** 
    - improved flipping the flag to opt-in.

  - **test coverage** 
    - added unit tests in `tests/cell/darkmode/` for `isColorAutomatic`, `getDarkModeCorrectedColor` and `_getKeepsAutomaticTextColorAsIs`, plus regression pins for the data-bar and theme-corruption bugs, all mutation-tested (broke each fix on purpose, watched the test fail, restored). 
    - **still missing print-path**, cause this one probably needs a much bigger fake setup, leaving it for a fast follow-up.
    - note: `591ca3151e` isn't the theme-corruption fix, the good one is `0f55894fb7`.


## Others

- ✅ **Done**
  - **leftover AI-translation comments** 
    - comments reworded to plain English, added "(translated from original Russian comment)" so we
don't lose that context.

  - **for...in on wsViews** 
    - `for...in` replaced by `.forEach`.

  - **async note on the color shuttle**
    - added to the JSDoc: not safe across an async boundary, a caller that awaits before reading it may see a different call's color.

- ❌ **Not found**
  - **isPrint/usePrintScale mix** 
    - checked this one closely, turns out `isPrint` (6097) and the `usePrintScale` line you flagged are in two separate functions rather than the same one reusing a var, so I didn't observe this issue. let me know if you were looking at something else though.

- ↪️ **Kept as follow up**
  - **ThemeColor identity coupling:** 
    - Looked into it, the fix looks small and low-risk (3 files, ~10 lines, no changes to clone()), but still need validation against a real file through the full pipeline, and goes beyond what this PR's diff needs to fix. I'll follow your final reco and keep it as a follow-up rather than expanding this PR's scope here.

  - **scope creep in common/skin.js:** 
      - agreed, pre-existing GlobalSkin bug bundled in.

  - **duplicated inFindResults call** 
    - agreed, not fixing here, keeping as a follow-up instead of bundling a refactor into this PR.

  - **_getKeepsAutomaticTextColorAsIs shape**
      - agreed, same as above, keeping as a follow-up instead of bundling a refactor into this PR.

---

Thanks again @MonaAghili , rest of the review checked out exactly as written. 
Would appreciate a re-review if you can 🙌

MonaAghili:
> **Note:** an earlier pass on this same branch reached an "Approve" verdict. This is a separate, deliberately hostile/adversarial re-review — treating the PR as guilty until proven safe, per a request to stress-test it against corruption, print, and accessibility risk. It surfaces gaps the earlier pass didn't flag, in particular the absence of a model-integrity round-trip test. Reconcile the two before deciding what actually blocks merge.

---

## Verdict

**REQUEST CHANGES**

---

## What the PR appears to fix

- Replaces the hardcoded white/light `defaultState.background`/`border` with values sourced from a new per-content-theme palette (`EditorSkins["theme-light"/"theme-dark"].CellBackground/CellGrid`), driven by `WorkbookView.updateDarkMode()` — this is the actual fix for "canvas defaults to white."
- Adds an opt-in `bIsFillRecolorable` flag to `drawFillCell` so only the search-highlight fill and the bare-canvas/no-fill background get dark-mode-corrected; explicit cell fills, data-bar colors, and CF/style/gradient preview swatches are explicitly excluded.
- Adds symmetric automatic-color correction for text (`StringRender`), decorations, and borders (`isColorAutomatic` + `getDarkModeCorrectedColor`), gated by `_getKeepsAutomaticTextColorAsIs` so default text isn't inverted on top of a light explicit fill.
- Fixes a real, previously-shipped corruption bug where `GlobalSkin.CellBackground/CellGrid` were aliased from the *interface* skin colormap, so switching UI theme silently stomped the *content* dark-mode cell colors; now split into `EditorSkins` (content axis) vs `GlobalSkin` (interface axis).
- Forces print/print-preview to a separate, permanently-light `cells.printState`, independent of `cells.defaultState`.

---

## Critical risks

### 1. `getDarkModeCorrectedColor` returns a shared mutable object, enforced only by convention
**Severity: High**

`DrawingContext.getDarkModeCorrectedColor` ([cell/graphics/DrawingContext.js:479-496](cell/graphics/DrawingContext.js#L479-L496)) returns `this._darkModeColorShuttle` — a single object overwritten on every call. Every one of its 5 call sites (`StringRender.js` ×3, `utils.js`'s `drawFillCell`, `WorksheetView.js`'s border draw) must consume it synchronously before any other caller invokes it again on the same `ctx`. This is documented in a JSDoc comment, not enforced.

**Failure scenario:** a future refactor that resolves a decoration color and a text color before drawing either (e.g. to dedupe `setFillStyle` calls) would silently make one overwrite the other. Nothing throws; underlines/strikethroughs render in the wrong color — the kind of bug that slips past visual QA because the two colors involved are usually close (both derived from the same automatic black/white pair).

**Evidence:** tests confirm the shuttle's documented overwrite behavior, but nothing guards against a future caller violating the contract.

### 2. Print-path isolation is asserted, not proven
**Severity: High**

Automatic-color correction is gated purely on `ctx.isDarkMode`. `WorkbookView.updateDarkMode()` only flips `isDarkMode` on `buffers.main`, `buffers.overlay`, and `cellEditor`'s contexts (`_getOwnedDrawingContexts`). The print-preview context (`printPreviewState.getCtx()`) is never touched by that method — it relies entirely on being constructed fresh with the `DrawingContext` constructor's default (`isDarkMode = false`) and never being reassigned.

**Failure scenario:** any code path that reuses an existing `DrawingContext` for a print/export job instead of constructing a fresh one would inherit `isDarkMode = true` from whatever screen context it was borrowed from — printed/exported output would get inverted automatic text/border colors on top of the correctly-light `printState.background`.

**Evidence:** no test under `tests/cell/darkmode/` touches `drawForPrint`, `printPreviewState`, or PDF export. Explicitly deferred by the author — but for an office suite, "printed/exported output doesn't match the document" is a first-class failure mode, not a fast-follow.

### 3. No round-trip test proving dark-mode rendering leaves the document model untouched
**Severity: Critical**

Manual tracing of every `getDarkModeCorrectedColor` and `drawFillCell` call site found none that mutate model objects — `getDarkModeCorrectedColor` only reads primitives (`getR()/getG()/getB()/getA()`) and never writes back into a `Fill`/`RgbColor`/`ThemeColor` owned by the model. That conclusion comes from manual inspection, not from anything the test suite demonstrates.

**Failure scenario:** a future "cache the corrected fill on the cell for perf" change stores a corrected color back onto `c.getFill()` or `font.getColor()`. Since no test opens a real workbook, toggles dark mode, renders, and diffs the model's colors (or serializes and compares bytes) against a light-mode baseline, such a regression would pass the entire existing suite undetected.

**Evidence:** every test in `DarkModeTests.js` operates on hand-built fakes (`makeFakeDrawingContext`, `makeFakeWorksheetView`, `makeFakeFillCtx`). None loads a real `Workbook`/`Worksheet`, none checks cell color identity/value before vs. after a dark-mode render pass, none serializes and reloads. **This is not demonstrated.**

### 4. `isColorAutomatic`'s `ThemeColor` branch depends on a global singleton identity
**Severity: Medium**

`isColorAutomatic` ([cell/model/WorkbookElems.js:492-500](cell/model/WorkbookElems.js#L492-L500)) treats a color as automatic either via an `isAutoColor` flag (survives `.clone()`, tested) *or* via `color === g_oDefaultFormat.Font.c` reference identity — valid only as long as nothing reassigns `g_oDefaultFormat.Font.c` to a new `ThemeColor` instance mid-session without every existing cell's captured reference staying in sync.

**Failure scenario:** an actual Office theme change (not a dark-mode toggle) that legitimately creates a new default-text `ThemeColor` would make cells whose font color was captured before the swap silently stop being recognized as "automatic" — they'd stop getting dark-mode-corrected until reload. Correctness regression, not corruption, but exactly the "tests confirm the mechanism, not the real lifecycle" pattern this review was asked to be suspicious of.

**Evidence:** the existing test manually swaps `defaultFormat.Font` and restores it — it never exercises the real theme-change code path that would reassign this in production.

---

## Test suite assessment

**Convincingly covered**
- `isColorAutomatic`: null/undefined, flagged vs. literal-black, clone preservation, `ThemeColor` identity — correctly tests identity (`strictEqual`), not just value equality.
- `getDarkModeCorrectedColor`: shared-shuttle identity, RGB cache keying, overwrite on a different-color call, alpha pass-through.
- `_getKeepsAutomaticTextColorAsIs`: light/dark solid fills, no-fill, gradient/pattern with and without a resolved fallback, and search-highlight overriding the fill result in both directions (proves short-circuit, not coincidence).
- `drawFillCell`'s opt-in guard: omitted-flag, explicit opt-in, light-mode no-op — directly pins the data-bar regression class from commit `42705a02f4`.
- Theme-corruption regression: targeted test proving `canvas-cell-background`/`canvas-cell-grid` no longer write into `GlobalSkin.CellBackground/CellGrid`, with a negative control proving `updateGlobalSkin` isn't just a no-op.

**Not demonstrated**
- Model-integrity round-trip: that a real cell/workbook's fill/font-color objects are bit-for-bit unchanged after a dark-mode render pass (see Risk 3).
- `WorkbookView.updateDarkMode` end-to-end: that toggling dark mode updates every owned context, and that repeated light→dark→light→dark toggling is idempotent with no stale cached value.
- Multiple worksheets/views sharing one workbook staying consistent on a dark-mode toggle.
- Contrast at/near the `isColorDark`'s `L < 100` threshold boundary — only far-interior points (black/near-black vs. white/yellow) are tested.
- Data bar or CF highlight color legibility against a dark-mode-corrected background (tested in isolation from each other, never together).
- Any print/export path (author-acknowledged gap).
- That `getDarkModeCorrectedColor`'s shuttle-reuse contract can't be silently violated by two call sites racing within one synchronous draw call — only that the overwrite behavior *exists*.

---

## Questions the author must answer before merge

1. Does `printPreviewState.getCtx()`'s `DrawingContext` ever get constructed by copying/reusing a context from `buffers.main`/`overlay`/`cellEditor`, or is it always a fresh instance whose `isDarkMode` defaults to `false` and is never subsequently set? Point to the exact construction site.
2. Is `RgbColor.prototype.isAutoColor` ever read/written through a JSON-based path (clipboard's internal format, co-editing change objects, undo/redo `getProperty`/`setProperty`) rather than only direct property access at render time? Could it leak into a collaborative-editing payload and be misinterpreted by a peer with different `g_oDefaultFormat` state?
3. `_getKeepsAutomaticTextColorAsIs` checks contrast against the cell's own **uncorrected** fill, since explicit fills are never recolored. Confirm no rendering path (print/PDF included) ever recolors an explicit fill outside `drawFillCell`, which would desync that assumption.
4. For every current call site of `getDarkModeCorrectedColor`, show the exact statement that consumes the returned value before any other code could call it again on the same `ctx`. Has this been checked specifically for `_drawCellCFDataBar`/`_drawCellCF`'s iteration over conditional-formatting rules?
5. Does `_cleanCellsTextMetricsCache()` (called on every worksheet on dark-mode toggle) hold any pre-computed *colors*, as opposed to purely layout/metrics data? If not, what actually invalidates any color computed and cached elsewhere across repeated toggling in one session?

---

## Recommended follow-up tests

1. **Model-integrity round-trip** — render a fixture workbook (mixed explicit fills, automatic text, `ThemeColor` fonts, data bars, CF highlights) once in light mode and once in dark mode; assert every color-bearing model object is unchanged in both cases.
2. **Print isolation smoke test** — force `isDarkMode = true`, run the real print-preview/print path, assert every color the print renderer's context reports is unaffected.
3. **Toggle idempotency** — light → dark → light → dark; assert `defaultState.background/border` and a fixed cell's rendered color return to bit-identical values each time.
4. **Multi-worksheet consistency** — two+ `wsViews` on one `WorkbookView`; toggle once; assert all views render consistently.
5. **CF/data-bar contrast on dark background** — assert data bar and CF highlight colors remain distinguishable against a dark-mode-corrected default background, not just "stays unmodified."
6. **Shuttle contract guard** — a debug-mode assertion or static check ensuring every call site consumes `getDarkModeCorrectedColor`'s result immediately, so a future violation fails loudly instead of silently swapping colors.

---

## Final recommendation

Real diagnostic work went into this: the theme-corruption fix (`GlobalSkin`/`EditorSkins` aliasing) is a legitimate, well-understood bug with a targeted regression test, and the opt-in `drawFillCell` guard is the right shape of fix for "don't touch explicit colors." `getDarkModeCorrectedColor`/`isColorAutomatic` don't mutate model objects anywhere I could find, and print output is structurally routed around a separate `printState`.

But "structurally routed around" is not "proven isolated," and for a spreadsheet engine handling millions of documents, the burden of proof for "this cannot silently alter a saved or printed document" sits with the PR. Before merge: add a model-integrity round-trip test (Risk 3), add at least one test that exercises the real print path rather than relying on `isDarkMode` defaulting to false and never leaking (Risk 2), and get a real answer on the `g_oDefaultFormat.Font.c` identity fragility under an actual theme-change lifecycle (Risk 4) rather than the current hand-constructed test. The color-shuttle non-purity (Risk 1) can ship as a documented, contained tradeoff — but it should be tracked, not silently accepted.

MonaAghili (changes requested):
Did a deeper pass on the automatic-vs-explicit color algorithm itself, not just the latest diff, since the author flagged they're reconsidering it after testing AI → Google Sheets → XLS → EuroOffice imports.

**The core classification is fine; the gap is that it's all-or-nothing with no fallback.** `isColorAutomatic` (`cell/model/WorkbookElems.js:492-500`) and the `isTextRecolorable` gate in `cell/view/StringRender.js:1481` (and its duplicate at ~1554) correct a color only if it's flagged automatic; anything else is trusted as intentional and gets zero contrast check, no matter how dark the background it ends up on. That's a reasonable default, but it means any place upstream that silently strips the "automatic" bit from a color that's still functionally default text produces permanently unreadable black-on-dark output — dark mode can't self-correct because the model no longer says it's allowed to.

Traced a concrete case where that stripping actually happens in this repo, not just hypothetically: `ReadColorSpreadsheet2` (`cell/model/Serialize.js:2117-2128`) reads a color record off the binary stream and only branches on `color.theme` / `color.rgb` — it never looks at `color.auto`, even though `ReadColorSpreadsheet` (`common/SerializeCommonWordExcel.js:634-635`) does populate that field from the stream. So a color that comes in flagged `auto=true` with an `rgb` fallback (the standard way a converter round-trips "this was automatic" through a legacy format like XLS — `<color auto="1" rgb="000000"/>`) gets turned into a plain `RgbColor(0,0,0)` with `isAutoColor` unset. From there:

- cell has no fill → `_getKeepsAutomaticTextColorAsIs` (`WorksheetView.js:6507-6508`) correctly says the background needs correction
- but `isColorAutomatic(textColor)` returns `false` for that RGB(0,0,0) → `isTextRecolorable` is `false` → the text is left black
- rendered against the dark canvas background → invisible

`tests/cell/darkmode/DarkModeTests.js:48` actually asserts this exact classification as correct ("explicit RgbColor(0,0,0) is NOT automatic — literal black is not the auto default"), which is true as a statement about the in-memory model, but it locks in the assumption without checking whether the resulting render is readable — so the suite can't catch this.

I don't think this argues for the alternative you were weighing (always force dark text light on a dark background regardless of source) — that would break intentionally dark-on-dark branding/labels and CF-driven dark text, and I checked that conditional-formatting fills already do flow into `c.getFill()` via `sheetMergedStyles` before the contrast check runs, so that part of the design is sound. The fix is narrower: stop discarding `color.auto` in `ReadColorSpreadsheet2` — when it's true, produce an auto-flagged color even if an `rgb` fallback is present, instead of a plain `RgbColor`. That closes the actual leak at the point where the semantic information is lost, without touching the render-time algorithm or reopening the explicit-color-preservation guarantee.

Everything else I went looking for in the current WIP — ThemeColor identity lifecycle, toggle idempotency, pattern/gradient fill handling, multi-worksheet consistency — held up under a second look; no new issues there.

PR #79, opened 2026-08-25
fix: Use correct quotation marks for Danish

hhjuhl:
Danish typographic tradition uses two standards for quotation marks.

1. Guillemets pointing to the quote: »Example«.
2. Down-9s and Up-6s: „Example“.

Eurooffice currently has dual Up-9s set for Danish: ”Example”.

This commit changes Danish typographer's quotes or smart quotes to guillemets, which is the most widespread in professional Danish typography. 

I have intentionally left the single quotes as they are, since using the correct ones tends to cause issues with apostrophes.

### DIFF
--- a/word/Editor/Paragraph/Run/RunAutoCorrect.js
+++ b/word/Editor/Paragraph/Run/RunAutoCorrect.js
@@ -382,7 +382,7 @@
 		return true;
 	};
 	/**
-	 * Производим автозамену для умных кавычек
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):
Thanks for the pull request, this makes sense 👍

juliusknorr:
CI failure will be covered by https://github.com/Euro-Office/sdkjs/pull/92

PR #81, opened 2026-08-27
fix(build): emit ASCII-only bundles so doctrenderer's no-ICU V8 can parse them

### LINKED ISSUE #80: Terser emits bare astral object keys in sdk-all.js; doctrenderer's no-ICU V8 rejects them and x2t aborts (9.3.4 regression)
# Terser emits bare astral object keys in `sdk-all.js`; doctrenderer's no-ICU V8 rejects them and x2t aborts (9.3.4 regression)

**Target repo:** `Euro-Office/sdkjs`
**Labels (suggested):** bug, regression, build, priority:critical

---

## Summary

Since the Grunt → webpack/Terser migration (#65, shipped in Euro-Office 9.3.4), the
built `sdk-all.js` bundles contain **131 unquoted supplementary-plane ("astral")
object keys**, e.g. `𝕒:"\\doublea"` in the LaTeX symbol table. Terser treats those
characters as valid identifiers and drops the quotes.

`doctrenderer`/`x2t` embed **V8 8.9 built with `v8_enable_i18n_support=false`**.
Without ICU, V8's reduced Unicode tables do not classify astral characters as
`ID_Start`, so parsing the bundle fails with `SyntaxError: Invalid or unexpected
token`. `ScriptCompiler::Compile` then returns an empty `MaybeLocal`, and an
unguarded `.ToLocalChecked()` turns that into a `CHECK` failure →
`__builtin_trap()` → `Illegal instruction (core dumped)`.

The previous Closure Compiler build emitted **pure ASCII** (everything escaped to
`\uXXXX`), so this could never happen. The 9.3.2 bundle contains **zero** non-ASCII
bytes; the 9.3.4 bundle contains 43,348.

User-facing report: Euro-Office/DocumentServer#339 and
<https://help.nextcloud.com/t/euro-office-upgrade-to-9-3-4-from-9-3-3-failed/248589>

> ⚠️ Euro-Office/DocumentServer#339 attributes the crash to **Intel MPX**. That is
> incorrect — see "Ruled out" below. The `Illegal instruction` is V8 trapping
> itself, not an unsupported CPU instruction.

## Impact

- Euro-Office **9.3.4 is unusable**: `.deb` postinst, standalone image and
  cluster-docs all fail during `documentserver-generate-allfonts.sh`, and every
  document save/convert fails (callback status 7).
- Affects **amd64 and arm64** alike. Downgrading to 9.3.3 resolves it.
- Affected bundles: `word/sdk-all.js` and `slide/sdk-all.js` (131 bare astral keys,
  1,256 astral characters each).

## Root cause chain

1. Terser config in `build/webpack.sdk.factory.mjs` sets neither
   `format.ascii_only` nor `quote_keys`. Defaults are raw UTF-8 output and
   identifier-style (unquoted) object keys.
2. Output therefore contains `𝕒:"\\doublea"`, `𝔸:"\\doubleA"`, … while BMP
   characters stay quoted (`"∙":"\\bullet"`).
3. V8 is built without ICU:
   [`core/Common/3dParty/v8/tools/8.9/x64-linux-dynamic/nc-build.sh#L329`](https://github.com/Euro-Office/core/blob/03656316cf42d1f9ea4067be6e97d7ae9862c70a/Common/3dParty/v8/tools/8.9/x64-linux-dynamic/nc-build.sh#L329)
   (`arm64-linux-dynamic` is identical).
4. doctrenderer concatenates `sdk-all-min.js` + `fonts_*.js` + `sdk-all.js` and
   compiles the result as one script:
   [`core/DesktopEditor/doctrenderer/editors.cpp#L69-L135`](https://github.com/Euro-Office/core/blob/03656316cf42d1f9ea4067be6e97d7ae9862c70a/DesktopEditor/doctrenderer/editors.cpp#L69-L135)
5. The compile fails and is not checked:
   [`core/DesktopEditor/doctrenderer/js_internal/v8/v8_base.cpp#L95`](https://github.com/Euro-Office/core/blob/03656316cf42d1f9ea4067be6e97d7ae9862c70a/DesktopEditor/doctrenderer/js_internal/v8/v8_base.cpp#L95)
   — the only unguarded `ToLocalChecked()` on that path; its siblings at L61, L83
   and L112 all test `IsEmpty()` first.

## Minimal reproduction

Take the shipped 9.3.4 `word/sdk-all.js`, replace every non-ASCII byte with `x`
(known-good, identical byte length), and append **one line**:

```js
var __k={"𝕒":1};   // astral char inside a string literal → compiles fine
var __k={𝕒:1};     // same char as a bare identifier key  → V8 aborts
```

Oracle — inside `ghcr.io/euro-office/documentserver:v9.3.4`, a non-zero
`sdk-all.cache` means the script compiled:

```sh
DIR=/var/www/euro-office/documentserver
export LD_LIBRARY_PATH=$DIR/server/FileConverter/bin:$LD_LIBRARY_PATH
rm -f $DIR/sdkjs/*/sdk-all.cache
cd $DIR/server/FileConverter/bin && ./x2t -create-js-cache
stat -c %s $DIR/sdkjs/word/sdk-all.cache
```

## Evidence

| # | Variant | `word/sdk-all.cache` |
|---|---|---|
| 1 | stock 9.3.4 word bundle | **0 — abort** |
| 2 | 9.3.2 word bundle, same 9.3.4 x2t binary | 6,752,720 — OK |
| 3 | 9.3.4 `sdk-all-min.js` + 9.3.2 `sdk-all.js` | 6,752,720 — OK |
| 4 | 9.3.2 `sdk-all-min.js` + 9.3.4 `sdk-all.js` | **0 — abort** |
| 5 | 9.3.4 bundle, all non-ASCII → `x`, identical byte length | 5,331,872 — OK |
| 6 | restore 25 / 50 / 75 / 87 / 94 % of non-ASCII bytes | all OK |
| 7 | restore 99 % | **0 — abort** |
| 8 | restore **only** the math-table region (2,168 bytes) | **0 — abort** |
| 9 | ASCII bundle + `var __k={"𝕒":1};` | 5,331,952 — OK |
| 10 | ASCII bundle + `var __k={𝕒:1};` | **0 — abort** |

Rows 3/4 isolate the fault to the JS, not the binary. Row 5 rules out size.
Rows 9/10 are the 15-byte minimal case.

`node --check` passes on **every** variant above under Node 15 (V8 8.6), Node 16
(V8 9.4) and Node 22 (V8 12.4) — all ship full ICU. No Node- or Chromium-based
test can detect this.

## Ruled out

- **Intel MPX** (as claimed in Euro-Office/DocumentServer#339) — reproduced on
  arm64, where MPX does not exist.
- **ES-version skew / dropped Babel ES5 downlevel** — the bundles parse on V8 8.6,
  *older* than the embedded 8.9.
- **Bundle bloat** — 9.3.2's `sdk-all.js` is larger (19.7 MB vs 18.4 MB) and works.
- **Parser stack depth** — both parse at `--stack-size=600`.
- **One-byte vs two-byte string representation** — a single astral or Cyrillic
  character *inside a string literal* compiles fine.
- **Invalid UTF-8 / lexical corruption** — the file is valid UTF-8.
- **Leaked `@@license-banner@@` sentinel** — it sits inside `/* */`.
- **Stale code cache** — fresh installs fail too; the 0-byte cache is a consequence.

## Proposed fix (this repo)

In `build/webpack.sdk.factory.mjs`, `terserOptions`:

```js
format: {
    comments: /@@license-banner@@/,
    ascii_only: true,     // escape all non-ASCII as \uXXXX (restores Closure's invariant)
},
quote_keys: true,         // force "𝕒": instead of 𝕒:
```

Either option alone is likely sufficient; together they are equivalent to the
configurations proven to compile (rows 5 and 9).

**Regression test:** assert the built bundles contain no byte > `0x7F`. Cheap,
exact, and encodes the invariant the Closure build used to provide implicitly.

## Follow-ups in other repos

- `Euro-Office/core` — defensive hardening, worth its own issue:
  - guard the `ToLocalChecked()` at `v8_base.cpp#L95` and surface the TryCatch
    message; a JS syntax error must not abort the process;
  - do not create `sdk-all.cache` until `CreateCodeCache()` succeeds — an abort
    currently leaves a **0-byte** cache that poisons every later run;
  - evaluate `v8_enable_i18n_support=true` (separate trade-off: binary size, build
    time). The fix above does not depend on it.
- **CI gap** — this repo's `check-build.yml` validates bundles with Node 20 and
  headless Chromium, both full-ICU, so it structurally cannot catch this class of
  bug. The bundle needs to be parsed by the actual embedded V8. The three
  doctrenderer suites tracked in Euro-Office/core#93 are exactly that gate;
  Euro-Office/core#134 clears the last currently-unblocked suite.

## Related

- Euro-Office/DocumentServer#339 — user-facing report (title needs correcting)
- #65 — the webpack migration that introduced this
- #78 — separate webpack-migration bundle-integrity bug
- #73 — open Terser bump; whatever lands here must keep `ascii_only`/`quote_keys`

chrip:
Terser was configured with neither format.ascii_only nor format.quote_keys, so the webpack pipeline emitted the LaTeX symbol table's supplementary-plane ("astral") keys as bare identifiers -- 131 of them per bundle in word/ and slide/ sdk-all.js, alongside 43348 raw non-ASCII bytes.

doctrenderer/x2t execute these bundles on a V8 built with v8_enable_i18n_support=false (core: Common/3dParty/v8/tools/8.9/*/nc-build.sh). Without ICU its reduced Unicode tables do not classify astral characters as ID_Start, so it rejects the entire bundle with "SyntaxError: Invalid or unexpected token". ScriptCompiler::Compile then returns an empty MaybeLocal and an unguarded ToLocalChecked() aborts the process -- the "Fatal error in v8::ToLocalChecked / Empty MaybeLocal" and "Illegal instruction (core dumped)" that make Euro-Office 9.3.4 fail on first start and on every document save.

The Closure Compiler build this replaced escaped all non-ASCII to \uXXXX, so the ASCII-only invariant held implicitly. The webpack migration dropped it silently, and no gate noticed: every JS check in check-build.yml runs on Node and headless Chromium, both full-ICU, which parse these bundles happily.

- format.ascii_only restores the invariant.
- format.quote_keys is redundant with ascii_only today but pins the behaviour against a future Terser bump. It must sit inside `format`; Terser rejects a top-level quote_keys outright with "`quote_keys` is not a supported option".
- Two unit tests assert the options and that Terser escapes astral keys rather than emitting them bare.
- A CI step asserts the emitted bundles contain no byte > 0x7F, which is the only gate here capable of catching this class of regression.

Verified against the v9.3.4 release image: all eight bundles rebuild with zero non-ASCII bytes and zero bare astral keys, `x2t -create-js-cache` exits 0 with all four caches written (previously aborted leaving a 0-byte cache), and documentserver-generate-allfonts.sh completes cleanly.

Fixes #80

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -152,23 +152,14 @@ jobs:
       # bundle with "SyntaxError: Invalid or unexpected token", aborting x2t on
       # first start (issue #80). Every JS gate in this workflow runs on Node and
       # headless Chromium -- both full-ICU -- so none of them can catch it.
-      # ASCII-only output is the invariant the previous Closure build provided
-      # implicitly; assert it directly on the emitted bundles.
-      - name: Assert built bundles are ASCII-only (#80)
+      #
+      # The check is a script rather than a grep because the residual failure
+      # mode is pure ASCII: with format.ascii_only but no format.quote_keys,
+      # Terser emits `\u{1d552}:"\\doublea"`, which no byte-range test flags.
+      - name: Assert built bundles are ASCII-only with no bare astral keys (#80)
         run: |
           cd sdkjs
-          status=0
-          for f in deploy/sdkjs/*/sdk-all.js deploy/sdkjs/*/sdk-all-min.js; do
-            [ -e "$f" ] || continue
-            if LC_ALL=C grep -qP '[^\x00-\x7F]' "$f"; then
-              echo "::error file=$f::non-ASCII bytes in built bundle; Terser format.ascii_only must stay enabled (see issue #80)"
-              LC_ALL=C grep -oP '[^\x00-\x7F]+' "$f" | sort -u | head -5
-              status=1
-            else
-              echo "ok: $f is ASCII-only"
-            fi
-          done
-          exit $status
+          node build/scripts/check-bundle-ascii.cjs
 
       - name: Install QUnit runner dependencies
         run: |
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -145,6 +145,31 @@ jobs:
           npm ci --prefix build
           npm run --prefix build build
 
+      # doctrenderer/x2t execute these bundles on a V8 built with
+      # v8_enable_i18n_support=false, whose reduced Unicode tables reject
+      # supplementary-plane characters used as identifiers. Terser emitted the
+      # LaTeX symbol table's astral keys bare, and that V8 then failed the whole
+      # bundle with "SyntaxError: Invalid or unexpected token", aborting x2t on
+      # first start (issue #80). Every JS gate in this workflow runs on Node and
+      # headless Chromium -- both full-ICU -- so none of them can catch it.
+      # ASCII-only output is the invariant the previous Closure build provided
+      # implicitly; assert it directly on the emitted bundles.
+      - name: Assert built bundles are ASCII-only (#80)
+        run: |
+          cd sdkjs
+          status=0
+          for f in deploy/sdkjs/*/sdk-all.js deploy/sdkjs/*/sdk-all-min.js; do
+            [ -e "$f" ] || continue
+            if LC_ALL=C grep -qP '[^\x00-\x7F]' "$f"; then
+              echo "::error file=$f::non-ASCII bytes in built bundle; Terser format.ascii_only must stay enabled (see issue #80)"
+              LC_ALL=C grep -oP '[^\x00-\x7F]+' "$f" | sort -u | head -5
+              status=1
+            else
+              echo "ok: $f is ASCII-only"
+            fi
+          done
+          exit $status
+
       - name: Install QUnit runner dependencies
         run: |
           sudo apt-get update
--- a/build/scripts/check-bundle-ascii.cjs
+++ b/build/scripts/check-bundle-ascii.cjs
@@ -1,0 +1,103 @@
+/**
+ * SPDX-FileCopyrightText: 2026 Euro-Office contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+/**
+ * Guards the two invariants the built sdkjs bundles must satisfy for
+ * doctrenderer/x2t, which run them on a V8 built with
+ * v8_enable_i18n_support=false (core: Common/3dParty/v8/tools/8.9/*).
+ *
+ *   1. No raw non-ASCII byte.
+ *   2. No object key that is a bare supplementary-plane ("astral") identifier.
+ *
+ * (2) is the subtle one and the reason a plain "is it ASCII" grep is not
+ * enough: with format.ascii_only but without format.quote_keys, Terser emits
+ *
+ *     \u{1d552}:"\\doublea"
+ *
+ * which contains no byte > 0x7F yet still parses as a bare identifier whose
+ * code point is U+1D552. A no-ICU V8 rejects that exactly as it rejects the
+ * raw UTF-8 spelling, so invariant (1) alone would let issue #80 regress
+ * silently. See build/webpack.sdk.factory.mjs.
+ *
+ * Usage: node build/scripts/check-bundle-ascii.cjs [file...]
+ * With no arguments, scans the built sdk-all bundles under the deploy root.
+ */
+
+'use strict';
+
+const fs   = require('fs');
+const path = require('path');
+const { sync: globSync } = require('glob');
+const { resolveBuildRoot } = require('../lib/env.cjs');
+
+// An unquoted object key in `{`/`,` position that starts with either a raw
+// astral character or a unicode escape. A quoted key cannot match: the quote
+// sits between the delimiter and the escape.
+const BARE_ASTRAL_RAW = /[{,]\s*[\u{10000}-\u{10FFFF}]/u;
+const BARE_ASTRAL_ESC = /[{,]\s*\\u(?:\{[0-9a-fA-F]{5,6}\}|[dD][89abAB][0-9a-fA-F]{2})/;
+
+function sampleAround(text, regex, limit = 3) {
+    const global = new RegExp(regex.source, regex.flags.includes('g') ? regex.flags : regex.flags + 'g');
+    const out = [];
+    let m;
+    while ((m = global.exec(text)) !== null && out.length < limit) {
+        out.push(text.slice(m.index, Math.min(m.index + 40, text.length)));
+    }
+    return out;
+}
+
+function checkFile(file) {
+    const buf  = fs.readFileSync(file);
+    const problems = [];
+
+    let nonAscii = 0;
+    for (const byte of buf) if (byte > 0x7f) nonAscii++;
+    if (nonAscii > 0) {
+        problems.push(`${nonAscii} raw non-ASCII byte(s); Terser format.ascii_only must stay enabled`);
+    }
+
+    const text = buf.toString('utf8');
+    for (const [label, re] of [['raw', BARE_ASTRAL_RAW], ['escaped', BARE_ASTRAL_ESC]]) {
+        if (re.test(text)) {
+            const samples = sampleAround(text, re).map((s) => JSON.stringify(s)).join(', ');
+            problems.push(`bare astral object key (${label} form); Terser format.quote_keys must stay enabled -- e.g. ${samples}`);
+        }
+    }
+
+    return problems;
+}
+
+function main(argv) {
+    let files = argv.slice(2);
+    if (files.length === 0) {
+        const buildDir  = path.resolve(__dirname, '..');
+        const deployRoot = resolveBuildRoot(buildDir);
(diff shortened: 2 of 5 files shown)


moodyjmz:
## Review

**TL;DR:** Approve with reservations. The shipped fix is correct — I reproduced the build with the exact pinned toolchain (`terser@5.49.0`, `terser-webpack-plugin@5.6.1`, `webpack@5.108.4`) and confirmed `ascii_only: true` + `quote_keys: true` together produce fully-quoted output, resolving #80. But the PR's own reasoning about *why* it works is backwards, and as a direct consequence the new behavioral regression test can't fail and the CI gate can't detect the specific failure mode it's meant to guard against.

<details>
<summary>Findings</summary>

**1. The code comment has the mechanism backwards.** `build/webpack.sdk.factory.mjs:314-320` claims `ascii_only` alone "forces astral keys to be quoted and escaped" and calls `quote_keys` "redundant." Built through the actual pinned webpack+terser-webpack-plugin+terser pipeline:
```
ascii_only only        → \u{1d538}:"\\doubleA"    ← BARE identifier (ASCII text, but unquoted)
ascii_only+quote_keys  → "\u{1d538}":"\\doubleA"  ← quoted, as shipped
```
`quote_keys` is the load-bearing fix, not decoration. `\u{1d538}` as a bare IdentifierName still resolves to the astral code point for ID_Start classification — a no-ICU V8 rejects it the same as the raw UTF-8 byte. The comment, if trusted later, invites deleting the option that's actually doing the work.

**2. The regression test that's supposed to catch this can never fail.** `build/test/webpack-sdk-terser-options.test.cjs:135-141` calls `terser.minify(src, terserOptions)` directly, bypassing terser-webpack-plugin's `ecma` injection entirely. Ran it standalone: with or without `quote_keys`, the output comes back quoted, because terser's default (no `ecma` set) behaves conservatively. So this test would still pass even if `quote_keys` were deleted from the real config tomorrow.

**3. CI's grep gate is blind to this exact residual mode.** `.github/workflows/check-build.yml:161-166` greps for bytes > `0x7F`. `\u{1d538}:` is pure ASCII text — confirmed `grep -P '[^\x00-\x7F]'` reports clean on it. The config-presence unit test (`terserOptions.format.quote_keys === true`, lines 118-126) *does* correctly catch an outright deletion of the option, so this isn't a hole today — but the intended behavioral coverage is not exercising the real failure mode.

**4. A second, untouched Terser pass exists.** `build/scripts/deploy-assets.cjs:86-91` minifies `Native/*.js`, `libfont/engine/*` (the `fonts_*.js` files #80 itself names as concatenated by doctrenderer), `spell/*`, `hash/*`, `zlib/engine/*` — no `ascii_only`/`quote_keys`, outside the CI glob. ASCII-clean today only incidentally (the sole non-ASCII content is Cyrillic in comments, stripped by `comments: false`). Not broken now, but nothing prevents it from becoming the next #80.

**5. Minor CI hardening.** `[ -e "$f" ] || continue` makes an unmatched glob a silent pass; `grep -qP`'s exit 2 (grep error) is indistinguishable from "no match."

**Checked and not problems:** `format` block is unconditional across module/platform; `quote_keys` correctly nested inside `format` (top-level is rejected by Terser); `terserOptionsOf()` reads the live config object; `require('terser')` resolves the same pinned version the build uses; CI glob/output-path/chunk-naming line up; `grep -P` is available on ubuntu-latest; dev builds skip Terser with no hazard.

</details>

**Suggestion before merge:** correct the comment at `webpack.sdk.factory.mjs:314-320`, and set `ecma: 2020` (or derive it from the same source terser-webpack-plugin uses) in the `terser.minify()` call at `webpack-sdk-terser-options.test.cjs:135` so the test actually exercises the code path it's named for.

moodyjmz:
Verified the follow-up commit (991547e) directly rather than taking the commit message on faith — pulled the diff, re-derived Terser's actual identifier-printing rule (`output.js`: the gate is `ecma < 2015`, a boolean threshold, not a specific version — so pinning `ecma: 2020` in the test is a safe proxy for whatever terser-webpack-plugin injects), and confirmed the `glob` pattern resolves against the real deploy layout at the pinned `glob@8.1.0`.

All four points from the review above are correctly addressed:

- The `format.ascii_only`/`quote_keys` comment in `webpack.sdk.factory.mjs` now correctly states `quote_keys` is the load-bearing option, with the exact `\u{1d552}:` failure mode spelled out.
- The behavioral test now pins `ecma: 2020` so it actually exercises the code path the real build takes, and adds a "guards the guard" test asserting the bare key *does* reappear if `quote_keys` is removed — so this can't quietly go blind again.
- The CI gate is now `build/scripts/check-bundle-ascii.cjs`, checking both raw non-ASCII and bare-astral-key in both raw and escaped form, sharing its regexes with the test file.
- `deploy-assets.cjs`'s second Terser pass over `Native/*.js`/`libfont/engine/fonts_*.js` now gets the same options, closing the latent gap.

Nice catch-and-fix cycle. Approving.

moodyjmz (approved):
Approving — see review comments above; the follow-up commit correctly addressed all findings.

PR #82, opened 2026-08-27
fix(build): debounce webpack watch mode to avoid mid-write reads (#78)

### LINKED ISSUE #78: webpack watch mode intermittently reads sdkjs source mid-write, breaking the concatenated bundle and increasing build times
## Issue (at least on my setup)

While running `npm run watch:cell` (probably affects word/slide/visio too), a save often triggers an error and/or two rebuilds instead of one. When it errors, it's always the first build that fails, with this output:

```
ERROR in ./dummy.js 927344:1
Module parse failed: Unexpected token (927344:1)
File was processed with these loaders:
 * ./loaders/sdk-concat.cjs
| }(window));
|
> })(window);
```

`sdk-all.js` drops to under 1KB on that build. The second rebuild fires right after and produces the correct 30MB+ bundle. Not a one-off, it happens on most saves. **At the end the resulting total build time is 4-6 seconds longer and affects the dev experience (from codebase edit to browser test loop duration).**

**Environment:** reproduced consistently on WSL2.

## Root cause

Checked with a raw `inotifywait -m` timestamp log on the edited file during a real save. One save, one `CLOSE_WRITE`, but 7 separate `MODIFY` events before it (VS Code flushes the write in chunks). webpack's default `watchOptions.aggregateTimeout` is 20ms (`node_modules/webpack/lib/Watching.js`), way too short to cover that whole write burst.

`sdk-concat.cjs` reads every file listed in `configs/cell.json` on each rebuild, not just the changed one, since the whole point of the loader is keeping ~400+ files in one shared scope. If webpack fires the rebuild before the OS finishes flushing the just-saved file, the loader reads a truncated copy of it. That breaks the brace/paren balance across the full concatenated module, hence the parse error at the tail end of the bundle instead of near the actual edit.

Normal webpack apps have small, separately-compiled modules, so a torn read on one file just fails that one module and gets silently retried next tick. Here the whole 30MB `sdk-all` chunk for cell is one module (**hope future architecture will allow code splitting for better builds and frontend performance**), so any single file's torn read takes the entire bundle down for that cycle.

## Proposed fix that worked on my setup

Raise `watchOptions.aggregateTimeout` in `webpack.sdk.factory.mjs` (currently unset, so it's on the 20ms default) to something like 500ms. Incremental rebuilds already run 4-8s, so 500ms extra debounce is a small tax against removing a reproducible race. Probably easy to solve through a tiny PR.

MonaAghili:
## Summary

- Fixes #78: `npm run watch:cell` (and `watch:word`/`watch:slide`/`watch:visio`) could intermittently produce a truncated `sdk-all.js` (<1KB) with a `Module parse failed` error, immediately followed by a correct rebuild.
- Root cause: webpack's default `watchOptions.aggregateTimeout` is 20ms (confirmed in the installed `webpack@5.108.4` source). Editors can flush a single save as multiple `MODIFY`/rename events, especially over a WSL2/9p filesystem boundary. If a rebuild starts inside that 20ms window, `sdk-concat.cjs` re-reads the file mid-write and concatenates a truncated fragment into the ~400+ file bundle, producing a syntax error at the tail of the generated module.
- Fix: set `watchOptions.aggregateTimeout` to `500` in `chunkConfig()` in `build/webpack.sdk.factory.mjs`, shared by all four SDK targets (word/cell/slide/visio) and both chunk variants (`min`/`all`). `watchOptions` is a documented no-op for one-shot builds, so `npm run build` / CI are unaffected.
- Added `WATCH_AGGREGATE_TIMEOUT` env override for tuning on a given machine/CI watcher, consistent with this file's other env-tunable knobs. Invalid input (non-numeric/negative) warns and falls back to the default instead of throwing, so a stray/typo'd env var can never break a plain build.

## Why 500ms

25x webpack's default, comfortably absorbing a multi-event editor flush, while staying small relative to this pipeline's warm incremental rebuild time (~5-8s dev). It's a debounce on the rebuild *trigger*, not a read-time stability check — `sdk-concat.cjs` still does a plain `readFile` once the timeout fires, so this narrows the race window rather than eliminating it structurally. A write that's still landing after 500ms of silence (very slow disk, network mount) could theoretically still race; a full fix would require content-stability polling or retry-on-parse-failure in the loader, which is out of scope for this issue.

## Testing

- `npm test --prefix build`: 45/45 passing, including new coverage for:
  - default `aggregateTimeout` (500) on both chunks, all targets share the one factory
  - `WATCH_AGGREGATE_TIMEOUT` override, including `0` (explicitly honored, not treated as unset)
  - invalid override input warns and falls back rather than throwing, verified via `sdkConfig()` itself (not just the parsing helper)
- Live repro: ran `npm run watch:cell`-equivalent, then rewrote a config-listed source file in two staggered chunks (5-10ms apart) to simulate a multi-event editor save. Before the fix this pattern reliably races; with it, exactly one rebuild fired and the resulting `sdk-all-min.js` was full-size and passed `node --check`.
- Full production build of the `cell` module confirmed unaffected: `sdk-all.js` emitted at expected size (~19.8 MiB minified).

## Not in scope

- Structural (read-time) fix in `sdk-concat.cjs` — discussed and deliberately deferred; see comment in `webpack.sdk.factory.mjs`.
- WSL2/9p `inotify` events not firing at all (a separate, unrelated failure mode `watchOptions.poll` would address) — not what #78 reports.

Close #78

### DIFF
--- a/build/test/webpack-sdk-factory.test.cjs
+++ b/build/test/webpack-sdk-factory.test.cjs
@@ -175,6 +175,126 @@ test('StripBundlePostprocessPlugin: strips webpack\'s bootstrap "use strict" fro
     }
 });
 
+test('resolveWatchAggregateTimeout: falls back to the default when unset', async () => {
+    const { resolveWatchAggregateTimeout } = await import(
+        url.pathToFileURL(path.join(__dirname, '..', 'webpack.sdk.factory.mjs'))
+    );
+    assert.equal(resolveWatchAggregateTimeout(undefined), 500);
+    assert.equal(resolveWatchAggregateTimeout(''), 500);
+});
+
+test('resolveWatchAggregateTimeout: honors an explicit 0 (no debounce) instead of treating it as unset', async () => {
+    const { resolveWatchAggregateTimeout } = await import(
+        url.pathToFileURL(path.join(__dirname, '..', 'webpack.sdk.factory.mjs'))
+    );
+    // `Number(x) || 500` would wrongly collapse "0" back to the default since
+    // 0 is falsy — this is the regression the helper guards against.
+    assert.equal(resolveWatchAggregateTimeout('0'), 0);
+});
+
+test('resolveWatchAggregateTimeout: rejects a non-numeric override instead of silently falling back', async () => {
+    const { resolveWatchAggregateTimeout } = await import(
+        url.pathToFileURL(path.join(__dirname, '..', 'webpack.sdk.factory.mjs'))
+    );
+    assert.throws(() => resolveWatchAggregateTimeout('not-a-number'), /must be a non-negative number/);
+    assert.throws(() => resolveWatchAggregateTimeout('-100'), /must be a non-negative number/);
+});
+
+test('sdkConfig: never throws on a bad WATCH_AGGREGATE_TIMEOUT — one-shot builds must not fail over a watch-only setting', async () => {
+    const { sdkConfig } = await import(
+        url.pathToFileURL(path.join(__dirname, '..', 'webpack.sdk.factory.mjs'))
+    );
+
+    const tmpRoot = fs.mkdtempSync(path.join(os.tmpdir(), 'sdk-watch-options-bad-env-test-'));
+    const prevBuildRoot = process.env.BUILD_ROOT;
+    const prevCacheDir  = process.env.WEBPACK_CACHE_DIR;
+    const prevTimeout   = process.env.WATCH_AGGREGATE_TIMEOUT;
+    const prevWarn      = console.warn;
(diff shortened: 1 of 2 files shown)


j-base64 (approved):
Good, this correctly addresses #78. The root cause matches what I found (webpack's default 20ms aggregateTimeout too short for a multi-event editor save), the fix raises the watch-mode debounce, and look solidly covered: unit tests for resolveWatchAggregateTimeout (default, explicit 0, invalid input) and sdkConfig (timeout applied, env override respected), plus the WATCH_AGGREGATE_TIMEOUT env var for anyone who needs to tune it further. 
**Approving 💯**

PR #84, opened 2026-09-03
fix(pdf): prevent PDF annotation date corruption on save

### LINKED ISSUE #357: PDF comment dates become 1970 with an invalid "=" timezone after the editing session is closed and the file is saved
Hi, thanks for Euro-Office.

When I add a comment in the PDF editor, the comments panel shows the real time (for example `9/3/26, 10:03 AM`). 

That date stays correct while the session is still open. 

After every editor is closed and Document Server sends the save callback (status 2, ~10 seconds later), 

the saved PDF has a bad date. Reopening the file shows 1/20/70.

## Expected

After the last client closes and the document is saved to storage, `/CreationDate` and `/M` on the annotation should still be the comment time, with a normal PDF timezone (`+`, `-`, or `Z`):

```
D:20260903100300+03'00'
```

## Actual

This only happens after session close + save. The written file contains:

```
/CreationDate (D:19700120194248=00'00')
/M (D:19700120194248=00'00')
```

Year 1970, and `=` instead of `+` / `-` / `Z`. Force-save while editors are still connected is not required; closing the last tab is enough.

This looks like a Unix timestamp in seconds passed to something that treats it as milliseconds (`new Date(seconds)` → ~20 Jan 1970).

## Environment

- Image: `ghcr.io/euro-office/documentserver:v9.3.4-hotfix.1` (latest GitHub release as of 2026-08-28)
- Fast co-editing + autosave enabled (PDF otherwise reports status 4 on close and never writes the comment)

## Repro

Minimal repo: [REPO_URL](https://github.com/mustafahlvc/euro-office-pdf-comment-date)

```
docker compose up --build
# wait until http://localhost:3000 is up and http://localhost:8081/healthcheck returns true
cd app && npm install && npm run test:comment-date
```

The Playwright test: add a comment → open the comments panel → close every editor (lobby) and wait for status 2 → reopen the comments panel. You should see `1/20/70` only after that save. The test fails with:

```
PDF comment date is corrupted after save. Expected a 2026 date with timezone + / - / Z, got: D:19700120…=00'00'
```

Manual steps are in the README.

mustafahlvc:
- Fix: PDF comment/annotation dates showed as 1/20/70 or similar after a save+reload.
Root cause: CChangesPDFAnnotCreationDate/ModDate stored the ms-epoch timestamp via a 32-bit WriteLong, silently truncating it (value mod 2^32) every time it round-tripped through the collaborative changes journal.

- Fix: formatTimestampToPDF writing an invalid = timezone marker for UTC (PDF spec only allows Z/+/-); now emits Z.

 Fixes Euro-Office/DocumentServer#357

### DIFF
--- a/pdf/src/annotations/base.js
+++ b/pdf/src/annotations/base.js
@@ -1773,7 +1773,8 @@
         // modification date
         if (Flags & (1 << 5)) {
             let sModDate = memory.GetString();
-            this.SetModDate(sModDate);
+            let oModDate = ParsePDFDate(sModDate);
+            this.SetModDate(oModDate ? oModDate.getTime().toString() : sModDate);
         }
     
         // user ID
@@ -1945,7 +1946,8 @@
             let CrDate = null;
             if (memory.annotFlags & (1 << 4)) {
                 CrDate = memory.GetString();
-                this.SetCreationDate(CrDate);
+                let oCrDate = ParsePDFDate(CrDate);
+                this.SetCreationDate(oCrDate ? oCrDate.getTime().toString() : CrDate);
             }
     
             let oRefTo = null;
--- a/pdf/src/annotations/base.js
+++ b/pdf/src/annotations/base.js
@@ -2000,7 +2000,7 @@
 
     function ParsePDFDate(sDate) {
         // Регулярное выражение для извлечения компонентов даты
-        let regex = /D:(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})([Z\+\-]?)(\d{2})?'?(\d{2})?/;
+        let regex = /D:(\d{4})(\d{2})(\d{2})(\d{2})(\d{2})(\d{2})([Z\+\-=]?)(\d{2})?'?(\d{2})?/;
 
         // Используем регулярное выражение для извлечения компонентов даты
         let match = sDate.match(regex);
@@ -2021,8 +2021,8 @@
             let date = new Date(Date.UTC(year, month - 1, day, hour, minute, second));
 
             // Учитываем смещение времени
-            if (timeZoneSign === 'Z') {
-                // Если указано "Z", это означает UTC
+            if (timeZoneSign === 'Z' || timeZoneSign === '=' || !timeZoneSign) {
+                // Если указано "Z" или "=", это означает UTC
             } else if (timeZoneSign === '+') {
                 date.setHours(date.getHours() - timeZoneOffsetHours);
                 date.setMinutes(date.getMinutes() - timeZoneOffsetMinutes);
@@ -2071,15 +2071,12 @@
         // Calculate timezone offset
         let timezoneOffsetMinutes = date.getTimezoneOffset();
         
-        let timezoneOffsetSign;
-        if (timezoneOffsetMinutes < 0)
-            timezoneOffsetSign = '+';
-        else if (timezoneOffsetMinutes > 0)
-            timezoneOffsetSign = '-';
-        else if (timezoneOffsetMinutes == 0)
-            timezoneOffsetSign = '=';
+        if (timezoneOffsetMinutes === 0)
+            return 'D:' + year + month + day + hours + minutes + seconds + 'Z';
 
-        
+        let timezoneOffsetSign = timezoneOffsetMinutes < 0 ? '+' : '-';
+
+
         let timezoneOffsetHours = Math.abs(Math.floor(timezoneOffsetMinutes / 60)) >> 0;
         if (timezoneOffsetHours < 10)
             timezoneOffsetHours = '0' + timezoneOffsetHours.toString();
--- a/pdf/src/history/annotsChanges.js
+++ b/pdf/src/history/annotsChanges.js
@@ -1110,13 +1110,13 @@ CChangesPDFAnnotContents.prototype.private_SetValue = function(Value)
 
 /**
  * @constructor
- * @extends {AscDFH.CChangesBaseLongProperty}
+ * @extends {AscDFH.CChangesBaseStringProperty}
  */
 function CChangesPDFAnnotCreationDate(Class, Old, New, Color)
 {
-	AscDFH.CChangesBaseLongProperty.call(this, Class, Old, New, Color);
+	AscDFH.CChangesBaseStringProperty.call(this, Class, Old, New, Color);
 }
-CChangesPDFAnnotCreationDate.prototype = Object.create(AscDFH.CChangesBaseLongProperty.prototype);
+CChangesPDFAnnotCreationDate.prototype = Object.create(AscDFH.CChangesBaseStringProperty.prototype);
 CChangesPDFAnnotCreationDate.prototype.constructor = CChangesPDFAnnotCreationDate;
 CChangesPDFAnnotCreationDate.prototype.Type = AscDFH.historyitem_Pdf_Annot_Creation_Date;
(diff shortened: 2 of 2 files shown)


DmySyz (changes requested):
Thank you for the contribution!

Claude review:

<pre><b>Findings</b>                                                                                                                                                                              
                                                                                                                                                                                      
  <b>1. HIGH — </b><font color="#B1B9F9"><b>ReadFromBinaryBase</b></font><b> and </b><font color="#B1B9F9"><b>ReadFromBinaryBase2</b></font><b> parse dates from the file format as raw strings but pass them straight to </b><font color="#B1B9F9"><b>SetModDate</b></font><b>/</b><font color="#B1B9F9"><b>SetCreationDate</b></font>                             
                                                                                                                                                                                      
  The PR description says it &quot;corrects parsing in <font color="#B1B9F9">ReadFromBinaryBase(2)</font>&quot; by wrapping the value through <font color="#B1B9F9">ParsePDFDate().getTime().toString()</font>. But the current source on main at lines 1776
   and 1948 already passes the raw string directly. The diff summary says the PR adds:                                                                                                
                                                                                                                                                                                        
  <font color="#12488B">let</font> oModDate = ParsePDFDate(sModDate);                                                                                                                                                
  <font color="#2AA1B3">this</font>.SetModDate(oModDate ? oModDate.getTime().toString() : sModDate);                                                                                                                 
                                                                                                                                                                                        
  This converts a PDF date string like <font color="#B1B9F9">&quot;D:20260917120000Z&quot;</font> into a millisecond timestamp string like <font color="#B1B9F9">&quot;1789747200000&quot;</font>. That&apos;s fine for new annotations — but what about <b>existing </b>         
  <b>annotations already stored as PDF date strings</b>?                                                                                                                                       
                                                                                                                                                                                        
  Before this PR, <font color="#B1B9F9">_modDate</font> and <font color="#B1B9F9">_creationDate</font> hold PDF date strings. After this PR, they hold millisecond-epoch strings. <font color="#B1B9F9">formatTimestampToPDF</font> calls <font color="#B1B9F9">parseInt(timestamp)</font> — it works on    
  epoch strings but silently returns <font color="#B1B9F9">Invalid Date</font> on a PDF date string that starts with <font color="#B1B9F9">&quot;D:&quot;</font>.
                                                                                                                                                                                        
  This means annotations loaded from an old collaborative session (whose change journals still hold PDF date strings via the old <font color="#B1B9F9">CChangesBaseLongProperty</font>) will be read back correctly  
  by the new <font color="#B1B9F9">CChangesBaseStringProperty</font> (both store strings on the wire), but the <b>interpretation</b> changes: old values are PDF date strings, new values are epoch strings, and nothing
  distinguishes them. <font color="#B1B9F9">formatTimestampToPDF(oldPdfDateString)</font> produces garbage.                                                                                                          
                                                                                                                                                                                      
  <b>Fix</b>: <font color="#B1B9F9">SetModDate</font>/<font color="#B1B9F9">SetCreationDate</font> (or the callers) should normalize: if the input matches the <font color="#B1B9F9">D:</font> prefix pattern, parse it to epoch; if it&apos;s already a numeric string, keep it. Or add a 
  migration in the <font color="#B1B9F9">private_SetValue</font> that detects and converts.
                                                                                                                                                                                        
  <b>2. HIGH — </b><font color="#B1B9F9"><b>&quot;=&quot;</b></font><b> backward compatibility in </b><font color="#B1B9F9"><b>ParsePDFDate</b></font><b> is incomplete</b>                                                                                                                    
   
  The PR adds <font color="#B1B9F9">=</font> to the regex <font color="#B1B9F9">[Z\+\-=]?</font> so old files that contain <font color="#B1B9F9">&quot;=&quot;</font> as the timezone marker can still be parsed. But <font color="#B1B9F9">&quot;=&quot;</font> is then not handled in the <font color="#B1B9F9">if/else if</font> chain — it falls through 
  to neither <font color="#B1B9F9">Z</font>, <font color="#B1B9F9">+</font>, nor <font color="#B1B9F9">-</font>, so the timezone offset is silently ignored (the date is treated as UTC, which happens to be correct for <font color="#B1B9F9">&quot;=&quot;</font> since it was intended to mean UTC).             
                                                                                                                                                                                        
  The PR description says this is handled (&quot;Expanded timezone handling to treat <font color="#B1B9F9">&apos;=&apos;</font> and missing timezone values as UTC equivalents to <font color="#B1B9F9">&apos;Z&apos;</font>&quot;). The diff summary confirms a new branch at  
  line 2024. Good — but verify the actual implementation handles the fall-through explicitly rather than relying on implicit &quot;no-op means UTC&quot;. An explicit <font color="#B1B9F9">else if (timeZoneSign === </font>
  <font color="#B1B9F9">&apos;=&apos; || !timeZoneSign)</font> with a comment would be clearer and wouldn&apos;t break if someone adds a default branch later.                                                                      
                                                                                                                                                                                      
  <b>3. MODERATE — </b><font color="#B1B9F9"><b>formatTimestampToPDF</b></font><b> still uses local time, not UTC</b>                                                                                                                     
   
  Lines 2062-2067: <font color="#B1B9F9">date.getFullYear()</font>, <font color="#B1B9F9">date.getMonth()</font>, <font color="#B1B9F9">date.getHours()</font>, etc. — these are all local-time accessors. The timezone offset calculation at line 2070                        
  (<font color="#B1B9F9">date.getTimezoneOffset()</font>) compensates, so the <i>value</i> is correct. But the server-side converter (x2t, DoctRenderer) runs in the container&apos;s timezone (usually UTC), while the browser
  runs in the user&apos;s timezone. Two users in different timezones co-editing the same PDF will produce different date strings for the same millisecond epoch — they represent the same    
  instant, but the string bytes differ, causing unnecessary collaborative changes journal traffic.                                                                                    

  Not a blocker (it&apos;s pre-existing behavior, not introduced by this PR), but worth noting for a follow-up: use <font color="#B1B9F9">getUTCFullYear()</font> etc. and always emit <font color="#B1B9F9">Z</font>.                                 
   
  <b>4. MODERATE — the JSDoc </b><font color="#B1B9F9"><b>@extends</b></font><b> comments are wrong after the fix</b>                                                                                                                     
                                                                                                                                                                                      
  Lines 1113 and 1130 still say <font color="#B1B9F9">@extends {AscDFH.CChangesBaseLongProperty}</font> while the actual prototype chain is changed to <font color="#B1B9F9">CChangesBaseStringProperty</font>. This will mislead anyone reading  
  the code or using IDE inference.                                                                                                                                                    
                                                                                                                                                                                        
  <b>Not addressed by the review (out of scope)</b>                                                                                                                                            
   
  - The <font color="#B1B9F9">m_sOOTime</font> values passed from comment data at lines 198-199 — what format are they in? If they&apos;re already epoch strings, the PR is fine. If they&apos;re PDF date strings, finding 1  
  applies to these too.</pre>


From my side I agree. You can ignore issue 3 as it is out-of-scope. Issue 1 would be the critical one for the approval, but dealing with others would be nice as well if you are willing.

mustafahlvc:
Hi, 

Fixed finding 1. created shared normalizeAnnotDate() helper function to detect and convert if necessary. It's called directly from  SetModDate/SetCreationDate.

DmySyz (approved):

DmySyz:
> Hi,
> 
> Fixed finding 1. created shared normalizeAnnotDate() helper function to detect and convert if necessary. It's called directly from SetModDate/SetCreationDate.

Hi, thanks! One last thing - could you add the sign-off to the commit message? You'll now it is correct once the DCO check passes.

PR #85, opened 2026-09-15
fix(word): preserve firstPageOnly flag through background copy and save

DmySyz:
- document-background.js: copy() now preserves firstPageOnly so the flag is not silently dropped during internal document copies (the root cause of background loss on save); draw() skips rendering on pages after the first when the flag is set; constructor initialises firstPageOnly to false
- Serialize2.js: ReadBackground handles c_oSerBackgroundType.FirstPageOnly (=3) and sets oBackground.firstPageOnly; WriteBackground emits the record on save; FirstPageOnly added to c_oSerBackgroundType enum
- Document.js: pass nPageIndex to Background.draw() so the per-page guard receives the actual page number

To be merged together with: https://github.com/Euro-Office/core/pull/142

Assisted-by: ClaudeCode:claude-sonnet-4-8


MonaAghili (commented):
## Summary

- The core fix is correct as far as it goes: `copy()`, `draw()`, the constructor default, and the `Serialize2.js` `FirstPageOnly` read/write are all internally consistent, symmetric, and backward-compatible.
- Page-index semantics are sound: `Background.draw()` has exactly **one** call site in the whole repo (`word/Editor/Document.js:5336`), inside `CDocument.prototype.Draw`, which indexes `this.Pages[nPageIndex]` — a zero-based array. Every rendering path (screen, print, print-preview, PDF/base64 export) funnels through this single choke point starting at index 0, so "page 0 = first page" is correct everywhere and no caller regresses.
- **But the PR does not close the state-loss path it claims to fix.** Three independent, verified ways still exist for `firstPageOnly` to be silently dropped after this PR merges — one of them (the `isDefault()` gate) loses the flag on a bare open→save with **zero user action**.
- No tests were added, and the repo has no existing precedent for binary-serialization round-trip tests to model from.
- Verdict: **changes required** — not because the diff is wrong, but because it's incomplete relative to its own stated goal.

## Findings

### [P0] Critical — `isDefault()` gates the entire Background write, and doesn't know about `firstPageOnly`

- **File:** `word/Editor/document/document-background.js:170-176` (`isDefault()`), gated at `word/Editor/Serialize2.js:5298-5300`
- **Issue:** `WriteDocument` only emits the Background record at all when `!Document.Background.isDefault()`:

  ```js
  if (oThis.Document.Background && !oThis.Document.Background.isDefault()) {
      this.bs.WriteItem(c_oSerParType.Background, function(){oThis.WriteBackground(oThis.Document.Background);});
  }
  ```

  `isDefault()` only inspects `Unifill`/`shape`/`Color`; it never checks `firstPageOnly`.
- **Why it's wrong:** `ReadBackground` sets `firstPageOnly = true` in its own independent branch, with no dependency on Color/Unifill/shape also being present. So a Background with `firstPageOnly === true` and no fill is a directly reachable state. On the very next save, `isDefault()` returns `true`, the whole `WriteItem` call is skipped, and `firstPageOnly` (along with everything else) is gone — **before the PR's own `WriteBackground`/`FirstPageOnly` logic ever runs.**
- **Repro:** Load a document whose Background block contains only the `FirstPageOnly` marker (no Color/ColorTheme/pptxDrawing). `Background.firstPageOnly` becomes `true`. Save with no edits. `isDefault()` → `true` → Background item never written. Reload: `firstPageOnly === false`.
- **Fix:** `isDefault()` should return `false` when `this.firstPageOnly` is true — e.g. add `if (this.firstPageOnly) return false;` at the top.

### [P1] High — `setBackgroundColor` unconditionally resets `firstPageOnly` to `false`

- **File:** `word/Editor/Document.js:5627-5633`; reached from `word/api.js:14668` (`asc_putPageColor`, the public "Page Color" API/UI action)
- **Issue:**

  ```js
  CDocument.prototype.setBackgroundColor = function(color, unifill)
  {
      let oldValue = this.Background.copy();
      let newValue = new AscWord.DocumentBackground(color, unifill, null);  // firstPageOnly defaults to false here
      AscCommon.History.Add(new CChangesDocumentPageColor(this, oldValue, newValue));
      this.Background = newValue;
  };
  ```

- **Why it's wrong:** `oldValue` correctly preserves `firstPageOnly` (thanks to the PR's fix to `copy()`), but `newValue` is built from scratch and never inherits it. There is no fixup anywhere else in the chain (`asc_putPageColor`, `CChangesDocumentPageColor.Undo`/`Redo`) — confirmed by exhaustive tracing.
- **Repro:** Load a document with `Background.firstPageOnly === true`. Call `Api.asc_putPageColor(color)` (ordinary Format → Page Color). `logicDocument.Background.firstPageOnly` is now `false`; the next save omits the `FirstPageOnly` record.
- **Fix:** preserve the flag explicitly, e.g. `newValue.firstPageOnly = this.Background.firstPageOnly;` before `History.Add`.

### [P1/P2] High — history-item binary (de)serialization for page-color changes never handles `firstPageOnly`

- **File:** `word/Editor/document/document-background.js:109-159` (`writeToBinary`/`readFromBinary`), used by `word/Editor/DocumentChanges.js:1146-1180` (`CChangesDocumentPageColor`)
- **Issue:** This is a **separate** serialization path from `Serialize2.js`'s `WriteBackground`/`ReadBackground` — it's the per-object binary encode/decode for `AscDFH.CChangesBase` history items, used by `History.Add` (which calls `WriteToBinary` immediately, `word/Editor/History.js:466`) and consumed by collaborative-editing change transmission (`CollaborativeEditing.js` `Send_Changes`) and reception/recovery-replay (`CollaborativeEditingBase.js` `Apply_Data`, which does `new fChangesClass(Class); oChange.ReadFromBinary(Reader)`). `writeToBinary`/`readFromBinary` use a bitmask covering only Color/Unifill/shape — `firstPageOnly` isn't in it.
- **Why it's wrong:** any `CChangesDocumentPageColor` created while `Old`/`New` carries `firstPageOnly === true` loses that value the moment it crosses this binary boundary — i.e., for any other collaborator, or on crash-recovery changes-log replay. A subsequent Undo on the receiving/replaying side restores the wrong Background shape.
- **Repro:** Set `firstPageOnly = true`, trigger a page-color change (creating `Old.firstPageOnly = true`), serialize via `WriteToBinary`, reconstruct via a fresh instance's `ReadFromBinary`. Both `Old`/`New` come back with `firstPageOnly === false`.
- **Fix:** add a `firstPageOnly` bit to the same flags long in `writeToBinary`/`readFromBinary`, mirroring the `copy()`/`Serialize2.js` fixes.

### [P3] Design question — `firstPageOnly` is document-absolute, not per-section

- **File:** `word/Editor/document/document-background.js:47-50` vs. the existing `SectPr.Get_TitlePage()` / `Get_SectionPageNumInfo` "different first page" concept used for headers/footers
- **Issue:** `draw()`'s guard checks the document's absolute `pageIndex === 0`. In a multi-section document — e.g. `Get_MailMergedDocument` (`word/Editor/Document.js:17758`), which correctly `.copy()`s a `firstPageOnly=true` template background across a merge that inserts a section break per record — only the first page of the *first* record shows the background; the first page of record 2, 3, … does not, even though each is a "first page" in the same sense Word's own per-section header/footer model already recognizes.
- **Not necessarily a bug:** this may match Core's intended contract (page background as a document/master-page-level property, not per-section), which isn't reviewable here. Flagging as a concern to confirm with Core, not a required change.

### [P3] Pre-existing, not a regression — some document-clone paths never copy `.Background` at all

`Api.GetMailMergeTemplateDocContent` (`word/apiBuilder.js:~5456`) and the print-selection temp document (`word/Drawing/DrawingDocument.js:~7151`) never set `.Background` on the cloned document at all (not even Color/Unifill, let alone `firstPageOnly`). This predates PR #85 and isn't a regression it introduced — noted only because it's the same "Background silently absent" family the review swept for.

## firstPageOnly Lifecycle

```
Binary → ReadBackground → Background.firstPageOnly → copy() → draw(pageIndex) → WriteBackground → Binary
```

| Stage | Status |
|---|---|
| Binary → `ReadBackground` | ✅ Correct — sets `firstPageOnly=true` on the `FirstPageOnly` record, symmetric with write |
| `Background.firstPageOnly` (default/constructor) | ✅ Correct — defaults `false`, so old binaries without the record behave exactly as before |
| `copy()` | ✅ Correct — the one thing the PR set out to fix, verified true/false both preserved |
| `draw(pageIndex)` | ✅ Correct — `pageIndex===0` reliably means first page; guard is stateless, no caching bug, no off-by-one on 1-page docs |
| → `WriteBackground` | ❌ **Broken** — gated by `isDefault()`, which ignores `firstPageOnly` (P0 finding) |
| Other Background *replacement* paths (`setBackgroundColor`, history-item binary log) | ❌ **Broken** — bypass `copy()`/`WriteBackground` entirely (P1 findings) |

## Page Index Verification

`nPageIndex` is confirmed zero-based and document-absolute: `CDocument.prototype.Draw` uses it directly as `this.Pages[nPageIndex]` (a plain JS array), and every rendering entry point in the repo (`DrawPage`, screen repaint, print, print-preview, PDF/base64 export, watermark render) iterates starting at `i = 0`. `Background.draw()` has exactly one caller in the entire codebase, so the fix is applied uniformly with no risk of a caller being missed. The `(pageIndex || 0) > 0` guard defensively fails toward *drawing* (not hiding) on any falsy input, which is the safe default. The one caveat is semantic, not a bug: "first page" here means the document's absolute page 0, not "first page of the owning section" — see the P3 design note above.

## Serialization Verification

`ReadBackground`/`WriteBackground` are symmetric for the fields they handle: `c_oSerBackgroundType.FirstPageOnly = 3` doesn't collide with existing values, sub-records are self-describing (type + length-prefixed) so write order doesn't matter, and the `Null`-length "presence flag" pattern used for `FirstPageOnly` matches multiple existing precedents elsewhere in the file (e.g. `Serialize2.js:6306`). Backward compatibility on **load** is correct — an old binary without the record leaves the constructor's `false` default untouched, and unknown sub-record types fall through to the file's normal `ReadUnknown` handling with no exception risk. The break is on the **save** side and in a **second, separate serialization path** (the history-item binary log), both detailed in the findings above — those are the real asymmetries, not the `ReadBackground`/`WriteBackground` pair itself.

## Caller/Regression Check

Safe. `Background.draw()`'s new `pageIndex` parameter is threaded through a single choke point (`CDocument.prototype.Draw`), so no existing caller was left behind or needs updating — this was verified by enumerating every caller of `Draw`/`DrawPage` in the repo (screen, print, print-preview, renderer-to-base64/watermark export). No regression was found here.

## Test Coverage

PR #85 adds no tests, and none of the following exist today (the repo's only word-editor unit tests, `tests/word/unit-tests/paragraphContentPos.js` and `deleted-text-recovery.js`, use a plain QUnit style — there's no existing binary-serialization round-trip test to model from, so this would be new infrastructure):

1. `Background` defaults to `firstPageOnly = false`
2. `copy()` preserves `firstPageOnly = true`
3. `copy()` preserves `firstPageOnly = false`
4. `ReadBackground()` sets `firstPageOnly = true` on the `FirstPageOnly` record
5. `WriteBackground()` emits the record when `firstPageOnly = true`
6. `true` survives a write → read round trip
7. A missing record produces `firstPageOnly = false`
8. `draw()` renders on `pageIndex = 0`
9. `draw()` skips on `pageIndex > 0` when `firstPageOnly = true`
10. Existing non-first-page-only backgrounds still render on every page
11. **(missing, required):** a save→load round trip of a Background with `firstPageOnly=true` and no Color/Unifill/shape — this is exactly the P0 finding and needs a regression test
12. **(missing):** `setBackgroundColor`/`asc_putPageColor` preserves `firstPageOnly` across a color change — the P1 finding
13. **(missing):** `CChangesDocumentPageColor.WriteToBinary`/`ReadFromBinary` round-trips `firstPageOnly` — the third finding

## Merge Recommendation

**Changes required.** The `copy()`/`draw()`/`Serialize2.js` fix is correct and well-targeted for the specific bug described in the PR, and page-index handling is sound. But it leaves the same bug class open in three other places it didn't touch — most seriously, `isDefault()` gating means `firstPageOnly` can be lost on a plain open-and-save with no user interaction at all, which undermines the PR's own stated purpose. Recommend fixing `isDefault()` (P0) and `setBackgroundColor` (P1) before merge at minimum; the history-binary gap (P1/P2) should be fixed in the same PR or immediately after, since it's the same root cause. Add the round-trip tests listed above, especially #11 and #12.

word/Editor/Serialize2.js
```diff
@@ -11589,6 +11594,8 @@ function Binary_DocumentTableReader(doc, oReadResult, openParams, stream, curNot
 			res = this.ReadDrawing (type, length, null, oDrawing);
 			if(null != oDrawing.content.GraphicObj)
 				oBackground.shape = oDrawing.content.GraphicObj;
+		} else if(c_oSerBackgroundType.FirstPageOnly === type) {
+			oBackground.firstPageOnly = true;
```
moodyjmz, line 11598: Returning `ReadOk` here asserts the handler consumed its declared length; this
one consumes nothing. `Read2` reads a `Variable` length via the *signed*
`GetULongLE` (`SerializeCommonWordExcel.js:740-743`) and advances
arithmetically, so a negative length spins the loop. Pre-PR this fell through
to `ReadUnknown` and `Skip2` rejected the negative. Not a new hole — the
`Color` branch above has the same shape — but the guard is free.
```suggestion
			if (0 !== length)
				res = c_oSerConstants.ReadUnknown;
			else
				oBackground.firstPageOnly = true;
```

word/Editor/Serialize2.js
```diff
@@ -1079,7 +1079,8 @@ var c_oSerDocPr = {
 var c_oSerBackgroundType = {
 	Color: 0,
 	ColorTheme: 1,
-	pptxDrawing: 2
+	pptxDrawing: 2,
+	FirstPageOnly: 3
```
moodyjmz, line 1083: Fork-local record claiming the next free value in a format upstream also
extends. If upstream ever assigns 3, files written by either side are misread by
the other — and on the next rebase that lands as a silent data bug, not a
merge conflict. Either take a high fork-local value (200+) or leave a
reservation comment here and in core's `BinReaderWriterDefines.h`.

word/Editor/document/document-background.js
```diff
@@ -33,16 +33,21 @@
...
 	{
-		return new DocumentBackground(this.Color, this.Unifill, this.shape);
+		let copy = new DocumentBackground(this.Color, this.Unifill, this.shape);
+		copy.firstPageOnly = this.firstPageOnly;
+		return copy;
 	};
-	DocumentBackground.prototype.draw = function(graphics, sectPr, theme, colorMap)
+	DocumentBackground.prototype.draw = function(graphics, sectPr, theme, colorMap, pageIndex)
 	{
+		if (this.firstPageOnly && (pageIndex || 0) > 0)
```
moodyjmz, line 49: Guard is on the document-absolute page index. Mail merge copies the background
into the merged document (`Document.js:17758`), so only page 0 of the whole
merge gets it, not the first page of each record. Fine if core#142's contract is
"document-level master-page property" — just confirm that's the intent.
`SectPr.Get_TitlePage()` is the precedent if it should be per-section.

moodyjmz (changes requested):
**Request changes.** Two blockers (findings 1 and 2), plus a one-liner worth
taking alongside them (finding 3). Nothing security- or performance-related; the
diff is clean on both counts.

The three things this PR touches are right. `copy()`, the `draw()` guard, the
`Serialize2` read/write pair — all correct, and I verified the wire format
against core#142: value 3 and the `Null` length encoding match on both sides,
byte for byte. Page-index semantics are sound too; `Background.draw()` has
exactly one caller and `nPageIndex` is document-absolute and zero-based on every
render path.

The problem is what the PR didn't touch. `DocumentBackground` has **four**
independent places that decide whether a field survives, and this PR updated two
of them:

1. the file format (`Serialize2.js` read/write) — done
2. the write gate (`isDefault()`) — **missed**
3. the history-item binary (`writeToBinary`/`readFromBinary`, i.e. co-editing and
   crash recovery) — **missed**
4. every `new DocumentBackground(...)` call site — **missed one**

A new field on this object has to go into all four or it dies somewhere. Right
now it dies in three places, one of which is a co-editing divergence.

None of these are hard fixes — roughly six lines total.

**One ordering constraint, and it matters: findings 1 and 2 must land together.**
Today the forward direction happens to agree, because `setBackgroundColor` builds
`New` with the flag already cleared. Fix finding 2 alone and nothing changes —
`New` is still `false` everywhere. Fix finding 1 alone and `New` starts carrying
a flag the wire strips, which *introduces* a forward-direction co-editing
divergence that doesn't exist right now. Landing 1 without 2 makes co-editing
strictly worse than the status quo.

---

### 1. HIGH — history binary drops the flag, co-authors diverge on undo

`word/Editor/document/document-background.js:109-159`

`writeToBinary`/`readFromBinary` use a flags bitmask covering `Color` (1),
`Unifill` (2), `shape` (4). `firstPageOnly` isn't in it. This is the pair
`CChangesDocumentPageColor` serializes through
(`word/Editor/DocumentChanges.js:1164-1176`) — the co-editing change payload and
the crash-recovery changes log. It is *not* the file format.

The forward direction happens to agree (because of finding 2 below, `New` is
`false` on both ends). Undo is where it breaks: the author's `Old` came from
`copy()` and carries the flag; the peer's `Old` was rebuilt by `ReadFromBinary`
and doesn't. User sets a page colour, hits Ctrl+Z — their screen shows the
background on page 1, the co-author's shows it on every page. Next save wins.
No error, no signal.

```js
// writeToBinary, after the shape block
if (this.firstPageOnly)
	flags |= 8;

// readFromBinary — unconditional, so it's idempotent
this.firstPageOnly = !!(flags & 8);
```

Wire-safe both ways: bit 8 carries no payload, old readers ignore it and consume
zero extra bytes, and a new reader on an old change gets `false`, which is the
correct pre-PR value.

### 2. HIGH — `setBackgroundColor` clears the flag on every page-colour change

`word/Editor/Document.js:5627-5633`

`oldValue` preserves `firstPageOnly` — that's this PR's `copy()` fix working.
`newValue` is built fresh from the constructor, so it gets the `false` default
(`document-background.js:39`) and nothing downstream repairs it; `Undo`/`Redo`
just assign `Old`/`New` wholesale.

Reachable from the public API with no exotic input — `asc_putPageColor`
(`word/api.js:14670-14693`) hits this on every Format → Page Color action,
including the "no fill" reset at `word/api.js:14681`.

Open an ODT with a first-page-only cover background, nudge the page colour once,
save: the background is now on every page. That's the bug this PR is fixing,
reached through a different door.

```js
let newValue = new AscWord.DocumentBackground(color, unifill, null);
newValue.firstPageOnly = this.Background.firstPageOnly;
```

(If the intended semantic is "a page-colour edit deliberately means whole
document", say so and add a comment — but make it a decision, not a default.)

### 3. LOW — `isDefault()` doesn't know about `firstPageOnly`, and it gates the entire write

`word/Editor/document/document-background.js:170-176`, gated at
`word/Editor/Serialize2.js:5298`

If `isDefault()` returns true, `WriteDocument` skips the whole Background item,
so the new `WriteBackground` branch at `Serialize2.js:5332-5335` never runs.

The reachable case is a **white** background: `isDefault()` returns true for
`#ffffff` (and `IsEqualRGB` at `word/Editor/Styles.js:9504` ignores `Auto`). A
first-page-only white cover round-trips to `firstPageOnly = false` with zero user
interaction — open, save, reload, gone.

Low, not critical, and I want to be straight about why: in every case where
`isDefault()` is true the background is white-or-absent, so dropping the record
is visually a no-op *today*. It bites one step later — fix finding 2, then
open that white first-page cover, save, set a red page colour, and you get red on
every page. One line, worth taking.

```js
if (this.Unifill || this.shape || this.firstPageOnly)
	return false;
```

### 4. LOW — the commit message's root cause doesn't match the code

The body says `copy()` dropping `firstPageOnly` was "the root cause of background
loss on save". It can't have been — `firstPageOnly` doesn't exist before this PR
(`grep -rni firstpageonly` returns only the lines this diff adds), and the old
`copy()` already forwarded `Color`, `Unifill` and `shape`
(`document-background.js:43`).

Whatever you actually reproduced, that isn't the mechanism. Worth fixing the
message; an inaccurate root cause in the history is what sends the next person
down the wrong hole.

### 5. LOW / hardening — the new read branch consumes zero bytes and doesn't check `length`

`word/Editor/Serialize2.js:11597-11598`

Returning `ReadOk` in this reader asserts "I consumed exactly `nRealLen` bytes".
The branch consumes none. `Read2` reads a `Variable` length via `GetULongLE`,
which is **signed** (`common/SerializeCommonWordExcel.js:740-743` — the `<< 24`
turns `0x80000000` into `-2147483648`), then advances arithmetically
(`stCurPos += nRealLen + nCurPosShift`, `:559`). A crafted type-3 record with a
negative `Variable` length spins the loop for ~10^9 iterations. Tab hang; no
over-read, both accessors are bounds-checked.

Pre-PR, type 3 fell through to `ReadUnknown` and `Skip2` rejected the negative
outright (`common/SerializeCommonWordExcel.js:699-703`).

**Not a new vulnerability** — the same primitive is already reachable through the
existing `Color` branch (`Serialize2.js:11580-11581`), so this adds no exposure.
But types 0/1/2 all consume something; 3 is the first that doesn't, and the guard
is free:

```js
} else if(c_oSerBackgroundType.FirstPageOnly === type) {
	if (0 !== length)
		res = c_oSerConstants.ReadUnknown;
	else
		oBackground.firstPageOnly = true;
}
```

The systemic fix — reject negative lengths in `Read1`/`Read2`/`Read2Spreadsheet`
— is pre-existing shared plumbing and wants its own PR. Don't bolt it onto this one.

### 6. LOW — enum value 3 is a fork-local claim on a shared format

`word/Editor/Serialize2.js:1079-1084`

Taking the next free value in a shared `.bin` enum works until upstream assigns 3
to something else, at which point files written by either side are misread by the
other. That collision surfaces as a silent data bug on the next rebase rather
than as a merge conflict.

Not blocking. Either take a deliberately high fork-local value (200+) or leave a
reservation comment next to the enum in both repos so the rebase trips over it.

### 7. Question — is `firstPageOnly` document-level or section-level?

`document-background.js:49` guards on the document-absolute page index. Mail
merge copies the background into the merged document
(`word/Editor/Document.js:17758`), so with `copy()` now preserving the flag, only
the first page of the *whole merge* gets the background — not the first page of
each record.

That may be exactly right if core#142's contract is "master-page property,
document-level". Just confirm it's a decision rather than a side effect of
`copy()` starting to work. `SectPr.Get_TitlePage()` is the existing precedent if
per-section semantics are wanted.

---

### Tests

One regression test — page-colour change preserves `firstPageOnly` — would be
proportionate here. A full round-trip matrix would mean inventing serialization
test infrastructure this repo doesn't have, and that's a separate piece of work.

---

moodyjmz:
Thanks for the rework. All four state-loss paths are closed, and the zero-length guard is in.

**Code is ready. One open UX question for @marcoambrosini: what should Format → Page Color do on a document whose background is first-page-only?** Today the flag is kept: the new colour paints page 1 only, "No fill" keeps the flag, and nothing in the UI can clear it. Approval pending that decision.

<details><summary>Details</summary>

### Open question: Page Color scope

Documents imported from ODT with a cover page style now carry a background that paints page 1 only. The Page Color control ([`asc_putPageColor`](https://github.com/Euro-Office/sdkjs/blob/132ce4effef7dd0c8acc853867f3cbd047caf5b5/word/api.js#L14668)) gives no indication of scope.

| Option | Behaviour | Consequence |
|---|---|---|
| A. Page Color applies to all pages | Setting a colour or "No fill" clears the flag | After a recolour the document no longer has a distinct cover background. Matches what the control implies today. |
| B. Keep first-page-only (current code) | The new colour paints page 1 only; "No fill" keeps the flag | A dark page colour with light text, chosen for the whole document, renders light-on-white on pages 2+ in the editor, print and PDF. The flag can never be removed. |
| C. Expose scope in the UI | A "first page only" toggle next to Page Color | Needs web-apps work; separate PR. |

Implementation for A: drop [`Document.js:5631`](https://github.com/Euro-Office/sdkjs/blob/132ce4effef7dd0c8acc853867f3cbd047caf5b5/word/Editor/Document.js#L5631). The only callers are the three `asc_putPageColor` branches (`api.js:14681`, `:14686`, `:14688`). `oldValue` still carries the flag, so undo restores it, and bit 8 keeps co-editing peers consistent.

```js
CDocument.prototype.setBackgroundColor = function(color, unifill)
{
	let oldValue = this.Background.copy();
	let newValue = new AscWord.DocumentBackground(color, unifill, null);

	AscCommon.History.Add(new CChangesDocumentPageColor(this, oldValue, newValue));
	this.Background = newValue;
};
```

Whichever option is chosen, consider [`isDefault()`](https://github.com/Euro-Office/sdkjs/blob/132ce4effef7dd0c8acc853867f3cbd047caf5b5/word/Editor/document/document-background.js#L177): under B, a background that is only the flag, with nothing to draw, is still written to the file.

### Non-blocking

- **Enum value 3 is a fork-local claim on a shared format.** [`Serialize2.js:1083`](https://github.com/Euro-Office/sdkjs/blob/132ce4effef7dd0c8acc853867f3cbd047caf5b5/word/Editor/Serialize2.js#L1083) and core's `BinReaderWriterDefines.h`. Add a reservation comment in both so an upstream rebase notices.
- **The page index is document-absolute.** Mail merge ([`Document.js:17759`](https://github.com/Euro-Office/sdkjs/blob/132ce4effef7dd0c8acc853867f3cbd047caf5b5/word/Editor/Document.js#L17759)) copies the flag, so only page 1 of the whole merge gets the background, not page 1 of each record. Confirm that is the intended contract.

### Fixed

| Item | Status |
|---|---|
| `setBackgroundColor` dropped the flag | Fixed; scope question above |
| History bitmask missing the flag | Fixed (bit 8, both directions) |
| `isDefault()` ignored the flag | Fixed |
| Zero-length record not checked | Fixed (`length != 0` → `ReadUnknown`) |
| Commit message root-cause claim | Fixed |

Pairs with core#142, which still needs changes; merge both together.

</details>

PR #86, opened 2026-09-17
Fix: save as picture blob url

DmySyz:
asc_SaveDrawingAsPicture previously created an <a> element with a data:image/... href and clicked it. This triggers frame-src CSP violations in browsers that enforce data: URI restrictions.

Decode the base64 payload and pass the raw bytes to AscCommon.DownloadFileFromBytes(), which uses URL.createObjectURL() internally, producing a blob: URL that satisfies the CSP.

Assisted-by: ClaudeCode:claude-sonnet-4-8


MonaAghili (dismissed):
Reviewed the diff and surrounding code (getImageDataForSaving/getBase64Img/ShapeToImageConverter, AscCommon.Base64.decode, DownloadFileFromBytes). No blocking issues — the base64 payload here is always canvas-generated PNG/JPEG, so the data-URL parsing and mime handling are safe for this call site, and DownloadFileFromBytes/revokeObjectURL lifecycle is unchanged, well-established code.

Two non-blocking notes left in the PR discussion for follow-up (not required for merge):
- Could reuse the existing getImageFileFromDataURL() mime-extraction helper in editorscommon.js instead of re-deriving it inline.
- If src is ever empty (canvas.toDataURL failure), Base64.decode(undefined) will throw uncaught since callers don't wrap this in try/catch — worth a guard, though it's an edge case and strictly safer than the previous behavior (empty-href anchor click).

Approving.

MonaAghili (commented):
## Blocking issues

None. The 3-line change is correct for its actual call site.

## Non-blocking issues

1. **Reuses hand-rolled base64/data-URL parsing instead of the existing `getImageFileFromDataURL()` helper.**
   (`common/apiBase.js:1188`) `common/editorscommon.js:2472-2477` already implements "split data URL → extract
   mime → `AscCommon.Base64.decode()`" for exactly this purpose (dataURL → uploadable file), using a regex
   (`arr[0].match(/:(.*?);/)[1]`) rather than the `substring`/`indexOf(";base64")` assumption in the new code.
   It wouldn't have produced a `File`/mime string in exactly the shape needed here, but the mime-extraction
   logic could have been factored out and shared instead of re-deriving it inline. Minor duplication, not
   worth blocking on.

2. **Unhandled exception if `getImageDataForSaving` returns a truthy object with an empty `src`.**
   (`common/apiBase.js:1188`) `getSelectionImageData()` (`common/Drawings/CommonController.js:9827-9834`) can
   legitimately return `{src: "", ...}` when `canvas.toDataURL` throws (caught) or when `NATIVE_EDITOR_ENJINE`
   is set. In that case `sSrc.split(",")[1]` is `undefined`, and `AscCommon.Base64.decode(undefined)` throws
   (`undefined.length`) since `decode()` never null-checks its input. The old code degraded silently
   (`a.href = ""`, `a.click()`). The new code throws synchronously out of `asc_SaveDrawingAsPicture`, which
   none of the four `web-apps` callers (`DocumentHolderExt.js` in cell/word/slide/pdf editors) wrap in
   try/catch. Not a data-loss regression, but it turns a silent failure into a hard JS error surfaced from a
   context-menu click. Suggest guarding: `if (oImageData && oImageData.src) { ... }`.

## Things that look correct

* **CSP fix is real and effective.** `asc_SaveDrawingAsPicture` always forces `bForceAsDraw = true` and passes
  an explicit `sImageFormat` of `"image/png"` or `"image/jpeg"` (`common/apiBase.js:1178-1182`), which routes
  through `getImageDataFromSelection` → `getBase64Img(true, sImageFormat)` → `ShapeToImageConverter`, i.e.
  always a canvas `toDataURL(sImageFormat)` result. So `sSrc` at this call site is guaranteed (barring the
  empty-string case above) to be `data:image/png;base64,...` or `data:image/jpeg;base64,...` — never SVG or
  another scheme — which makes the `substring`/`indexOf(";base64")` parsing and the `sImageFormat` mime type
  passed to `DownloadFileFromBytes` both correct in practice.
* **`AscCommon.Base64.decode()` correctly handles standard padded base64** — it trims the output `Uint8Array`
  to the actual decoded byte count (`common/hash/hash/engine.js:156-162`) rather than over-allocating based on
  the base64 string length, so PNG/JPEG byte content round-trips exactly.
* **`DownloadFileFromBytes()` accepts a `Uint8Array` directly** (`common/editorscommon.js:2433-2469`) —
  `new Blob([data], {type: mime})` works fine with a typed array, matching how it's already called elsewhere
  (`cell/api.js`, `slide/api.js`, `word/api.js`, `visio/api.js`) with byte arrays.
* **`revokeObjectURL` timing is unchanged and pre-existing** — the 200ms `setTimeout` before
  `removeChild`/`revokeObjectURL` is shared, already-battle-tested code used by every other download path in
  the SDK; this PR doesn't touch or risk that lifecycle.
* **Filename/extension preserved identically** — `sExt`/`sName` computation is untouched by the diff, so the
  downloaded filename (`Picture.png` / `Picture.jpeg`) is unchanged from before.

## Testing

No existing tests cover `asc_SaveDrawingAsPicture`. A full download E2E test isn't practical in the unit
suite, but worth adding: a unit test around `AscCommon.Base64.decode(dataUrl.split(",")[1])` for a known small
PNG/JPEG data URL, asserting decoded byte length and known header bytes (`89 50 4E 47` for PNG, `FF D8` for
JPEG) — catches base64-decode regressions without a browser download harness.

DmySyz:
The code duplication is a temporary preference as the priority for this PR is to be non-invasive.
Additionally reviewed by Claude.
Manually tested in Firefox and Chrome for different editors.

Aiiaiiio (approved):

moodyjmz:
Post-merge review. The CSP fix itself is correct — `bForceAsDraw=true` guarantees a canvas-generated data URL at this call site, `Base64.decode` round-trips the bytes exactly, and `DownloadFileFromBytes` takes a `Uint8Array` fine. Two low-severity follow-ups, neither a data-loss risk.

**TL;DR**
1. The blob MIME can disagree with the bytes — `Picture.png` containing PNG bytes, tagged `image/jpeg`.
2. An empty `src` is now a silent no-op: click, nothing happens, no error.

<details>
<summary><b>1 — Blob MIME lies about the bytes on multi-select (low, silent, new to this PR)</b></summary>

`common/apiBase.js:1191` passes `sImageFormat` as the blob MIME, but derives `sExt` from the actual data URL one line up. Those disagree whenever the source format isn't the one requested — and there is such a path:

- `getImageDataForSaving` (`common/Drawings/CommonController.js:9845`) only forwards `sImageFormat` when **exactly one** object is selected. With 2+ it calls `getSelectionImageData()` with no arguments, and that function hardcodes `_canvas.toDataURL("image/png")` (`CommonController.js:9866`).
- So multi-select → "Save as picture → JPG" gives `sSrc = "data:image/png;base64,…"`, `sExt = "png"`, and `DownloadFileFromBytes(u8arr, "Picture.png", "image/jpeg")`.

Before this PR the anchor carried the data URL's own type, so bytes and Content-Type always agreed. Now they can't.

Fix — derive the MIME from the source rather than the request, matching the idiom already on the line above:

```js
let sMime = sSrc.substring("data:".length, sSrc.indexOf(";base64"));   // "image/png"
AscCommon.DownloadFileFromBytes(u8arr, sName, sMime);
```

This also stops papering over the bug underneath it, which predates this PR and is filed separately as #89.

</details>

<details>
<summary><b>2 — The empty-src guard traded a crash for a silent no-op (low severity, bad shape)</b></summary>

`common/apiBase.js:1185-1197`. When `sSrc === ""`, `oImageData` is still a truthy object, so `if(oImageData)` passes — then both the `if` and the `else if (sSrc)` fail and the function returns having done nothing at all. No file, no error, no message.

Reachable from both branches:
- `getSelectionImageData` sets `sImageUrl = ""` on a tainted-canvas `toDataURL` throw (`CommonController.js:9868`) and unconditionally under `NATIVE_EDITOR_ENJINE` (`:9871`).
- `getBase64Img` returns `""` for a Notes/NotesMaster parent (`common/Drawings/Format/GraphicObjectBase.js:3782`), for degenerate geometry (`:3791`), and in `ShapeToImageConverter`'s catch when there is no `RasterImageId` (`word/Drawing/ShapeDrawer.js:2716`).

Severity is genuinely low — `NATIVE_EDITOR_ENJINE` isn't the browser context this PR targets, and the tainted-canvas route additionally needs no `RasterImageId` to fall through. But the *shape* is the bad one: the previous crash was at least visible, and a success-shaped non-save isn't. Worth an error surface, or at minimum a `console.warn`, instead of a bare `else if (sSrc)`.

</details>

<details>
<summary><b>Verified correct / not open — don't re-raise</b></summary>

- @MonaAghili's crash note (`Base64.decode(undefined)` throwing on empty `src`) — addressed by the guard in cf848df; her reviews were against the earlier 12f4674.
- @MonaAghili's duplication note (reuse `getImageFileFromDataURL`) — explicitly declined as a deliberate non-invasive choice. Fair.
- The `else if` anchor fallback is **not** a finding. It's reachable — a tainted canvas falls back to `getFullImageSrc2()`, an http URL — and an http href doesn't trip the `data:` CSP. It also fixes the old filename garbage, where `substring(11, -1)` on a non-data URL yielded `Picture.https://doc`.
- `Base64.decode` trims correctly via `new Uint8Array(dst.buffer, 0, writeIndex)` (`common/hash/hash/engine.js:158-161`), and `new Blob([view])` honours byteOffset/byteLength, so the round-trip is exact. `split(",")[1]` is safe — comma isn't in the base64 alphabet.

</details>

One process note, no lecture attached: the commit carries `Assisted-by: ClaudeCode`, the PR discussion says "additionally reviewed by Claude", and the approving review on the final commit has an empty body. Both findings above are things a human reading the diff against `getSelectionImageData` would have hit. The gap is verification, not disclosure.

PR #90, opened 2026-09-19
Fix spreadsheet wheel scroll suppression on macOS in Chromium browsers

kitzler-walli:
Fixes the root cause reported in Euro-Office/DocumentServer#315.

## Summary

`EventsController.js`'s wheel-target listener (`this.widget`, i.e. `#editor_sdk`) selects the legacy `mousewheel` event on macOS. Two other listeners are already registered on that exact same element using the modern `wheel` event:

- the web-apps layer (`DocumentHolderExt.js`, for Ctrl/Cmd+wheel zoom)
- the vendored `jquery.mousewheel` plugin (used for scrollbar normalization), which auto-detects and prefers `wheel` whenever the browser supports it

Chromium suppresses dispatch of the legacy `mousewheel`/`DOMMouseScroll` event to a target that already has a `wheel` listener registered on it ([chromium issue 40453359](https://issues.chromium.org/issues/40453359)). Because two `wheel` listeners already exist on this element, the SDK's own `mousewheel`-based scroll handler never receives real wheel input at all on macOS in Chromium-family browsers — independent of the delta-accumulation theory originally suspected in #315.

This explains the reported browser-specific split: reproducible in Brave, not reproducible in Chrome/Safari on the same machine (they evidently differ in this exact legacy-fallback behavior), and not present in ONLYOFFICE's own DocumentServer, whose equivalent web-apps/vendor code registers `mousewheel` consistently everywhere, so there's no `wheel` listener on the element to trigger the suppression.

## Fix

Removes the `!AscCommon.AscBrowser.isMacOs &&` exclusion so the feature detection is just "does this browser support `wheel`" — completing the modernization that's already been applied to the other two listeners on this element, rather than reverting them. With all wheel-registrations on the element agreeing on `wheel`, there's no legacy-fallback dispatch involved at all.

## Testing

Verified against a live, running Document Server via Chrome DevTools Protocol (Playwright), in a fresh browser context (no cache):

- Before the fix, `#editor_sdk`'s registered listeners: `[wheel (DocumentHolderExt), wheel (DocumentHolderExt), mousewheel (SDK), mousewheel (SDK), mousewheel (SDK)]`.
- After the fix (this one-line change, nothing else): `[mousewheel (jquery.mousewheel), mousewheel (SDK), mousewheel (SDK)]` — the two competing `wheel` listeners are gone, matching what ONLYOFFICE's own DocumentServer looks like for the same element.
- Replayed real captured macOS trackpad wheel deltas (`deltaMode: 0`, `isTrusted: true`, values matching an actual Brave/trackpad capture) through Playwright's `mouse.wheel()`:
  - Before: `Worksheet.scrollVertical` called 0 times, view never moved.
  - After: `scrollVertical` called on every event, view scrolled correctly, matching the input direction/magnitude.
- Confirmed by the original reporter directly in Brave on macOS: scrolling now works.

I also tested an alternative fix (reverting the two `wheel` registrations in `web-apps/apps/spreadsheeteditor/main/app.js` and `DocumentHolderExt.js` back to `mousewheel`/`DOMMouseScroll`) — it also resolves the bug, but reverts modernization work on two files and touches a vendored third-party library. This one-line SDK-side fix resolves it with no such tradeoffs, so I'm proposing it as the primary fix.

Credit to `@tiran133` on the DocumentServer issue for the original diagnosis and for identifying and testing this exact fix already — this PR is just carrying it through as a proper patch with additional live verification.


DmySyz (changes requested):
Hi, thanks a lot for this! All good from my side, the only thing missing is the sign-off in the commit message.

juliusknorr:
I rebased to trigger the other jobs which should pass now

DmySyz:
@kitzler-walli could you do the sign-off fix by tomorrow evening? I'd like to build a new RC and would be nice to get this in. Sorry for the rush.

kitzler-walli:
Sign-off added (ebce9fc). Looks like the CI re-run for this commit is stuck pending manual approval: https://github.com/Euro-Office/sdkjs/actions/runs/35889356672 — could someone with write access approve it? Trying to get this in before the RC build, per @DmySyz's note above.

DmySyz (approved):

